PR #11, opened 2017-10-10
Skip-List implementation

lembergerth:
Add a [skip list](https://en.wikipedia.org/wiki/Skip_list) implementation with partial counts.

Provides a sorted data structure with a mean performance of
O(log n) for insertion
O(log n) for removal
O(log n) for contains
O(log n) for random access by rank/index
O(1) for first/last item in order


kfriedberger:
Copyright notice for CPAchecker? Please change!

kfriedberger:
Is your implementation thread-safe?
How does it perform compared to a (Set based on a) [ConcurrentSkipListMap](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentSkipListMap.html)?

lembergerth:
> Copyright notice for CPAchecker? Please change!

This was the default for the existing IntelliJ configuration - I changed it.

> Is your implementation thread-safe?

No.

> How does it perform compared to a (Set based on a) ConcurrentSkipListMap?

My experiments haven't finished yet, but ConcurrentSkipListMap does not provide
random access - otherwise, the collections have the same theoretical background.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
```diff
@@ -0,0 +1,31 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.sosy_lab.common.collect;
+
+import java.util.SortedSet;
+
+public interface OrderStatisticSet<T> extends SortedSet<T> {
```
PhilippWendler: The interface and the methods should have JavaDoc that explains the guarantees.

Why is the name `OrderStatisticSet`? Why not `SortedUniqueList` or something else?

src/org/sosy_lab/common/collect/OrderStatisticSet.java
```diff
@@ -0,0 +1,31 @@
...
+
+  T get(int pIndexOf);
+
+  T remove(int pIndexOf);
```
PhilippWendler: Having `remove(int)` is problematic for `OrderStatisticSet<Integer>`, because calling `remove` with an `Integer` could call both methods. Better rename these two methods.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
lembergerth: Should I really add performance guarantees to the interface?
Despite that, JavaDoc is obviously missing, of course. I added it.

* Why not `SortedUniqueList`?

Lists provide more methods for indexed access (e.g., to add a value at a specific position),
so I thought that any interface name with `List ` in it may be misleading.
In addition, it would be easily possible to add a *MultiSet and *Map interface and implementation with similar capabilities, so the `Unique` part would have to be removed
and the new interfaces would have to get different names, reducing the 'grouping' of such interfaces.

* Why `OrderStatistic`?

It's not used in any interface description yet, we could easily add `OrderStatisticMultiset` and `OrderStatisticMap` interfaces, and the [meaning](https://en.wikipedia.org/wiki/Order_statistic) is correct: The interface allows 1) to get the k-th smallest value of a set and 2) to determine the k for a given value.
In addition, a data structure [order statistic tree](https://en.wikipedia.org/wiki/Order_statistic_tree) already exists in literature that provides the same capabilities.

That said, I'm open for names that are more self-explanatory, I just couldn't find one yet.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
lembergerth: Good point, I renamed them to {get,remove}ByRank , so that they are still close to the get/remove methods and easily findable.

kfriedberger:
Feature request: Is your implementation `Serializable`?
Most of our data-structures can be serialized.

lembergerth:
Now it is.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
PhilippWendler: Ok, if the name is already well-known, let's use it.

For guarantees in JavaDoc, I meant all kinds of guarantees, but performance would also be interesting if possible and meaningful for the interface (we could look at `Set`, `List`, etc., if they provide performance guarantees). Otherwise only `SkipList` should document performance guarantees.

lembergerth: The collection interfaces don't provide performance guarantees. But could  you have a look at the new JavaDoc to check whether you miss any information, please?

src/org/sosy_lab/common/collect/OrderStatisticSet.java
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * In both cases, the used compare-method should be consistent with <code>equals</code>, i.e.,
+ * <code>compare(a, b) == 0  =&gt;  a.equals(b)</code>, so that the contract provided by {@link
+ * java.util.Set Set} is fulfilled. Implementations of this interface should still work if the used
+ * compare-method is not consistent with <code>equals</code>.
+ *
+ * <p>If the used compare-method is not consistent with <code>equals</code>, unequal (i.e.,
+ * different) elements that are equal according to the order can exist in a set. In this case,
+ * elements that are added first will have a higher rank than elements of the same order that are
+ * added later.
```
PhilippWendler: Are you sure you want to guarantee a specific behavior if `compareTo` and `equal` are inconsistent? Why not simply require that they are consistent?

Also, this is a contradiction to the paragraph above, where it simply says that implementations "should" work (but are not required to) if the mthods are inconsistent.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
```diff
@@ -0,0 +1,113 @@
...
+   * 0, the largest element in the set has rank == size - 1.
+   *
+   * <p>If the used compare-method is not consistent with <code>equals</code>, i.e., <code>
+   * compare(a, b)</code> does <b>not</b> imply <code>a.equals(b)</code>, elements that were added
+   * first will have a higher rank than elements equal by comparison that were added later.
+   *
+   * <p>Example:
+   *
+   * <pre>
+   *     * Element type T: Pair&lt;Integer, Integer&gt;
```
PhilippWendler: `Pair` is not a good example because it is not defined (who knows how it's `equals` method is implemented?).

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -0,0 +1,756 @@
...
+
+    private Node(@Nullable T pValue, int pLevel) {
+      value = pValue;
+      next = new ArrayList<>(Collections.nCopies(pLevel + 1, null));
+      prev = new ArrayList<>(Collections.nCopies(pLevel + 1, null));
+      inBetweenCount = new ArrayList<>(Collections.nCopies(pLevel + 1, 0));
+      maxLvl = pLevel;
+    }
+
+    @Nullable
```
PhilippWendler: Is this formatted with `google-java-format`? I prefer annotations for return types on the same line as the return type.

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -0,0 +1,756 @@
...
+ *
+ * <p>Based on William Pugh: Skip Lists. A Probabilistic Alternative to Balanced Trees, 1990.
+ *
+ * <p>By design, a skip list is a single linked list. To be able to implement a random-access add
+ * operation in a simple manner, we extended the skip list to be double linked.
+ *
+ * @param <T> type of the elements of this list
+ * @see java.util.concurrent.ConcurrentSkipListMap
+ */
+public class SkipList<T> implements OrderStatisticSet<T>, Serializable {
```
PhilippWendler: Should be `final`.

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -0,0 +1,756 @@
...
+
+  public SkipList(SortedSet<T> pSortedSet) {
+    this(pSortedSet.comparator());
+    boolean changed = addAll(pSortedSet);
+    assert changed;
+  }
+
+  @SuppressWarnings("unchecked")
+  public SkipList() {
+    comparator = (Comparator<? super T>) Comparator.naturalOrder();
```
PhilippWendler: The unchecked cast should be replaced with `Comparator.<? super T>naturalOrder()` (same above).

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -0,0 +1,756 @@
...
+  private static final double ONE_HALF_LOG = Math.log(0.5);
+  private static final Random randomGenerator = new Random();
+
+  private final Comparator<? super T> comparator;
+  private Node<T> head = createHead();
+  private Node<T> tail = head;
+  private int size = 0;
+
+  @SuppressWarnings("unchecked")
+  public SkipList(@Nullable Comparator<? super T> pComparator) {
```
PhilippWendler: I would suggest factory methods instead of public constructors (Guava-style).

`copyOf` makes it much more clear than `new SkipList(foo)` that this is not a list that contains `foo`. Adding `of` with varargs makes it easier to create lists of a few elements.

Having overloads with `Collection` and `SortedSet` that differ in their semantic regarding which comparator is used is problematic. What if you want a copy of a `SortedSet`, but with a natural order? What if you pass a `SortedSet` instance to the method that accept's a `Collection`? People could expect that it still takes the `SortedSet`'s comparator.

So it is better to use factory methods with distinctive names that make clear with order is used, e.g., `copyOfWithNaturalOrder`, `copyOfWithSameOrder` etc.

If this gets too complex because of too many combinations that we want, we could use a builder instead.

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -0,0 +1,756 @@
...
+  }
+
+  @Override
+  public void clear() {
+    head = createHead();
+    size = 0;
+  }
+
+  private Node<T> getNode(int pIndex) {
+    if (pIndex >= size) {
```
PhilippWendler: In cases like this, it also needs to be chcked whether`pIndex` is negative. There exists `Preconditions.checkElementIndex` precisely for this use case.

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -0,0 +1,756 @@
...
+      public void add(T pT) {
+        throw new UnsupportedOperationException(
+            "Adding a node at a specific position doesn't " + "make sense in a sorted list");
+      }
+    };
+  }
+
+  @Override
+  public String toString() {
+    return "SkipList: [" + "size: " + size + " ]";
```
PhilippWendler: The `toString` of collections is pretty much standardized, and should be the same here as well.

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -0,0 +1,756 @@
...
+  }
+
+  @Override
+  public Comparator<? super T> comparator() {
+    return comparator;
+  }
+
+  @Override
+  public Iterator<T> iterator() {
+    return new ListIterator<T>() {
```
PhilippWendler: Why implement `ListIterator` if the return type is just `Iterator` anyway?

We could change the return type to `ListIterator`, or expose an additional `listIterator` method like `List` has. But probably exposing a `ListIterator` is not a good idea because it would allow `add` and `set` at a given position.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
```diff
@@ -0,0 +1,113 @@
...
+ * compare-method is not consistent with <code>equals</code>.
+ *
+ * <p>If the used compare-method is not consistent with <code>equals</code>, unequal (i.e.,
+ * different) elements that are equal according to the order can exist in a set. In this case,
+ * elements that are added first will have a higher rank than elements of the same order that are
+ * added later.
+ *
+ * @param <T> the type of elements maintained by this set
+ */
+public interface OrderStatisticSet<T> extends SortedSet<T> {
```
PhilippWendler: Does it make sense for all `OrderStatisticSet`s to extend `NavigableSet`? Or does it at least make sense for `SkipList`?

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -0,0 +1,756 @@
...
+
+  private static class Node<T> implements Serializable {
+
+    private static final long serialVersionUID = 10045121457565238L;
+
+    // These lists are transient because they are manually serialized to prevent stack overflows due
+    // to recursive structure
+    private transient List<Node<T>> next;
+    private transient List<Node<T>> prev;
+    @Nullable private T value; // May be null, so be careful with serializing
```
PhilippWendler: If google-java-format does not enforce the current formatting, I would prefer `private @Nullable T value`, because semantically the `@Nullable` is a modifier for the type `T`.

Why does one need to be careful while serializing `null` values?

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -0,0 +1,756 @@
...
+import java.util.ListIterator;
+import java.util.NoSuchElementException;
+import java.util.Random;
+import java.util.Set;
+import java.util.SortedSet;
+import javax.annotation.Nullable;
+
+/**
+ * A skip list implementation. Similar to ordered trees, but in list form. Achieves time complexity
+ * similar to ordered trees using a probabilistic approach.
```
PhilippWendler, line 45: Does it allow `null` values? This should be documented.

Do we want it to allow `null` values?

src/org/sosy_lab/common/collect/SkipListTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.Test;
+
+public class SkipListTest {
+
+  private static class TestSkipListGenerator extends TestStringSortedSetGenerator {
+
+    @Override
+    protected SortedSet<String> create(String[] pIntegers) {
+      SkipList<String> list = new SkipList<>();
+      // noinspection ResultOfMethodCallIgnored
```
PhilippWendler: What is this? Why is it necessary?

src/org/sosy_lab/common/collect/SkipListTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public static junit.framework.Test suite() {
+    TestSuite suite =
+        SortedSetTestSuiteBuilder.using(new TestSkipListGenerator())
+            .named("SkipList Test Suite")
+            .withFeatures(
+                CollectionSize.ANY,
+                CollectionFeature.SUPPORTS_ADD,
+                CollectionFeature.SUPPORTS_REMOVE,
+                CollectionFeature.SUPPORTS_ITERATOR_REMOVE,
```
PhilippWendler: The above three can be replaced with `SetFeature.GENERAL_PURPOSE`.

Probably we also want `CollectionFeature.KNOWN_ORDER`, `CollectionFeature.ALLOWS_NULL_VALUES` (depending on what we decide), and `CollectionFeature.SERIALIZABLE_INCLUDING_VIEWS`.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
lembergerth: Requiring that the used `compare`-method (not just `compareTo`!) is consistent with `equals`  is a strong constraint on the possible uses of the interface.

Whenever you want to use some comparison criterion that is not consistent with `equals`,
you will have to work around this by maintaining collections of objects for each 'comparison value', manually.

Through this definition in the interface, we would allow implementing sets to do this for the user.

> Also, this is a contradiction to the paragraph above, where it simply says that implementations "should" work (but are not required to) if the mthods are inconsistent.

Thanks!

src/org/sosy_lab/common/collect/OrderStatisticSet.java
lembergerth: It definitely makes sense for the `SkipList`.
Order statistic tree implementations are also able to get the floor/ceil values easily,
but I'm not sure whether this property can be derived from the interface, in general.

Do  you have an opinion on that?

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: No it doesn't.

I think the use of `null` as a value has very few legitimation and would be a special case
regarding the order of elements, too.

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: Ok.

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: > If google-java-format does not enforce the current formatting

No it doesn't, your choice makes sense - I changed it.

> Why does one need to be careful while serializing null values?

The comment was an artifact of a poor approach of manual serialization where I read the value as an Object and cast it, I removed it.

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: Yes, unfortunately.

> I  prefer annotations for return types on the same line as the return type.

Me too. The way it is formatted, it appears to be an annotation for the whole method.
Is there an easy way to change this?

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: The SortedSet-interface states that implementing methods should provide these four constructors.
If you say that we can disregard that, I'd be happy to add factory methods.

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: Will this work?
`T` does not have to extend Comparable, but the type parameter of `Comparator.naturalOrder()` does have to.

There's actually a bigger, more general problem: if no comparator is given and `T` does not implement the Comparable-interface.
But with the two constructors creating `Comparator.naturalOrder()`, there is no way of telling because we do not have an object of Type `T` available.

Looking at the `TreeMap` implementation, they also do not handle this.

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: Thanks for the hint!

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: > We could change the return type to ListIterator, or expose an additional listIterator method like List has. But probably exposing a ListIterator is not a good idea because it would allow add and set at a given position.

Yes, I thought of both, mainly because I implemented the `List`-interface, at first,
but did choose not to do either because of the same reasons.

I just kept the implementation in case it is required at some point, but keeping code lying around is a bad practice, of course. I'll remove it.

src/org/sosy_lab/common/collect/SkipListTest.java
lembergerth: It removes unnecessary inspections in IntelliJ, I removed it (and another one)

src/org/sosy_lab/common/collect/OrderStatisticSet.java
PhilippWendler: `SortedSet` also requires consistency of `equals` and `compareTo`, at least for fulfilling the `Set` contract. We should use the same guarantees and requirements.

`SkipList` doesn't call `equals` at all, right? Then it will work anyway just like users expect it.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
PhilippWendler: I know less about these data structures than you, so it is up to you to decide whether this makes sense.

src/org/sosy_lab/common/collect/SkipList.java
PhilippWendler: For the above reasons, I think that the Guava style is superior.

src/org/sosy_lab/common/collect/SkipList.java
PhilippWendler: This is another argument for factory methods, for those that use the natural order you can enfore `T extends Comparable<T>` via the type parameter of the method. Then you don't need the unchecked cast.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
lembergerth: Yes, the `SortedSet` interface states:
> […] if the sorted set is to correctly implement the Set interface

So that's equal to our condition, isn't it?

>SkipList doesn't call equals at all, right? Then it will work anyway just like users expect it.

Yes it does. All operations that are specific to a given object (e.g., `remove`, `contains`) use the equals-method to ensure that the objects are not only equal by the given Comparator, but by their own equals-method.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
lembergerth: OK, so how about that notion:

Every collection that is able to get the rank of an element in an ordered set has to determine the position of that element in the list.
If it can do that, it can also determine where an element in an ordered set *should* be, and thus determine its neighbors.

Thus, every implementation of `OrderStatisticSet` should also be able to implement `NavigableSet` efficiently.

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -185,6 +185,14 @@ public SkipList() {
     comparator = (Comparator<? super T>) Comparator.naturalOrder();
   }
 
+  /**
+   * Use the given {@link Random} object for future probabilistic computations.
+   */
+  public void reinitialize(Random pRandom) {
+    Preconditions.checkNotNull(pRandom);
+    randomGenerator = pRandom;
+  }
```
lembergerth: I'd like to point to this method so that it's not overseen:
Do you mind such a method?
The point is to enable the use of a controlled Random-object (with a specific seed)
from outside.

PhilippWendler: First, if the `Random` instance is static as it was before, it should probably be a `ThreadLocalRandom`.

Second, which use cases would need changing the random-number generator during the lifetime of a `SkipList` object, and would not be possible if the random-number generator could only be specified at instantation time?

Third, which use cases actually need specifying the random-number generator?
I can only think of cases where users wish to make their code deterministic, e.g., for tests.
However, I see no reason not to make `SkipList` fully deterministic by default.

Thus, I suggest to give each `SkipList` instance its own `Random` instance created with a constant seed.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
PhilippWendler: > All operations that are specific to a given object (e.g., remove, contains) use the equals-method to ensure that the objects are not only equal by the given Comparator, but by their own equals-method.

This should not be done. Just use only the given order, and document that if `equals` and the given order is inconsistent, then the `Set` contract is not fulfilled. This is the same as all other `SortedSet` implementations I know do.

lembergerth: Oh ok, I can do that.

src/org/sosy_lab/common/collect/SkipList.java
PhilippWendler: No, we can't change this.

But I noticed that you have many methods that are package-private but could be `private`. If you add the `private` modifier before the `@Nullable`, google-java-format will format as desired.

src/org/sosy_lab/common/collect/SkipListTest.java
PhilippWendler: It is still there.

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -163,37 +166,86 @@ private void readObject(ObjectInputStream pIn) throws IOException, ClassNotFound
...
+  public static <T extends Comparable<T>> SkipList<T> of(T p1, T p2, T p3, T... pOtherElements) {
+    SkipList<T> list = new SkipList<>();
+    list.addAll(Lists.newArrayList(p1, p2, p3));
+    for (T v : pOtherElements) {
+      list.add(v);
+    }
+    return list;
+  }
+
+  public static <T extends Comparable<T>> SkipList<T> create() {
```
PhilippWendler: Why not name this `of` in the Guava style?

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -163,37 +166,86 @@ private void readObject(ObjectInputStream pIn) throws IOException, ClassNotFound
...
+  public static <T> SkipList<T> create(Comparator<? super T> pComparator) {
+    return new SkipList<>(pComparator);
+  }
+
+  public static <T> SkipList<T> create(
+      Iterable<? extends T> pCollection, Comparator<? super T> pComparator) {
+    return new SkipList<>(pCollection, pComparator);
+  }
+
+  public static <T extends Comparable<T>> SkipList<T> createWithNaturalOrder(
```
PhilippWendler: How about `copyOfWithNaturalOrder`?

src/org/sosy_lab/common/collect/SkipList.java
```diff
@@ -163,37 +166,86 @@ private void readObject(ObjectInputStream pIn) throws IOException, ClassNotFound
...
+      Iterable<? extends T> pCollection) {
+    return new SkipList<>(pCollection);
+  }
+
+  public static <T> SkipList<T> copyOf(SortedSet<T> pSortedSet) {
```
PhilippWendler: How about `copyOfWithSameOrder`?

Or do you think the behavior of this method is what most people would expect from a `copyOf` method with no further explanation?

I just noted that Guava's `ImmutableSortedSet` has the reverse naming: `copyOf` uses natural order, `copyOfSorted` uses the same order. This is short and easy, and I think would be nice to adopt.

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: I used the [Guava Wiki](https://github.com/google/guava/wiki/CollectionUtilitiesExplained#static-constructors) as reference,
which states that newer Collection classes use `create()`.

I also find it more speaking than an `of` without parameters, and it can be easily found.

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: I had that first, but 'copy' implies that it's copied - and is a collection really copied if the position of its elements could be completely changed?

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: Here I used `copyOf` because the elements and their order are both kept, so it is a real 'copy' (so also 'yes' to your second question).

But for factory methods, Guava actually seems to differentiate between `ImmutableXXX` and other collections (e.g., `TreeMultiset`). The second only provide `create(...)` methods.

So should we just align ourselves with the `ImmutableXXX` classes for all factory methods, to have a naming scheme consistent with Guava classes, since you mentioned that `of(…)` methods are useful?

src/org/sosy_lab/common/collect/SkipList.java
PhilippWendler: This wiki page uses `create()` only as an example and does not say that newer collection classes use `create()`. However, mutable collections might be a better place to look at for reference than immutable collections if ours is mutable. This would indicate to have `create()` and no other `of` methods.

src/org/sosy_lab/common/collect/SkipList.java
PhilippWendler: That's maybe a valid point, but on the other hand the `WithNaturalOrder` makes it pretty clear what happens. 

I don't think that `create(Iterable)` really conveys its semantics well when used: Is it really obvious that `SkipList.create(foos)` returns a collection that contains the same as `foos`?

So maybe we find a third alternative?

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: Personally, I don't have any use case where you would need a sorted set of only `n` elements,
with `n` being that small that you write it down by hand (except for testing).
Thus, I don't require the `of` methods, but maybe you have a good use case?

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: How about `create(Iterable)` to create with natural order,
and `createOfSorted(SortedSet)` or `createWithSameOrder(SortedSet)`?

Edit:
I think the behavior of `create` is not clear by the name itself, but through the Java context.
Collections of the JDK (e.g., `ArrayList(Collection)` constructor ) as well as of Guava (e.g., `HashMultiset.create(Iterable)` factory method)
create a collection object containing the elements of the given Collection/Iterable, not just a view of it.

PhilippWendler: I would agree. Is it clear that `createWithSameOrder()` also copies the content and not only the order?

src/org/sosy_lab/common/collect/SkipList.java
PhilippWendler: Then let's remove it.

src/org/sosy_lab/common/collect/SkipList.java
lembergerth: Good point.

I would prefer `copyOfWithSameOrder`, similar to what I used before, but that would be even less consistent with the Guava convention.

Even though it is less than optimal, I think we can also assume that the parameter implies that the elements of the given list will be copied (otherwise, why not use a `Comparator`-parameter)?
But maybe you have some other good idea for the naming.

lembergerth:
Experiments have shown that a naive implementation of the `OrderStatisticsSet` with a `TreeMap` and `Iterables#get` shows performance comparable to the SkipList implementation, for waitlists the size we encounter them in sv-benchmark tasks. So for now, this implementation is unnecessary.

PR #14, opened 2017-10-26
Log manager interface extension

### LINKED ISSUE #12: LogManager: Support format strings for user and debug exceptions
Instead of using a simple string message in logDebugException/logUserException,
a formatstring with Object... args would be quite convenient.
This could easily be archived by overloading the method.

There could also be a benefit in perforamance if these string only get built if acutally logged. (see logf)

BCSol:
Implementation proposal to resolve #12.

Introduced `log( ... , Object... args)` and `logf(... , String format, Object... args)` for:
- `logUserException`
- `logDebugException`
- `logException`

Removed `logException(Throwable e)` since it's kind of the same as calling `logException(Throwable e,  Object... args)` with no additional arguments.

Dropped `@Nullable String` parameter altogether.
For old `log(.... , null)` calls, it will cause a compiler warning

### DIFF
--- a/src/org/sosy_lab/common/log/BasicLogManager.java
+++ b/src/org/sosy_lab/common/log/BasicLogManager.java
@@ -328,13 +328,12 @@ public class BasicLogManager implements LogManager, AutoCloseable {
    */
   @Override
   public void log(Level priority, Object... args) {
-    checkNotNull(args);
-
+    checkBuildAdditionalMessageParams(args);
     // Since some toString() methods may be rather costly, only log if the level is
     // sufficiently high.
     if (wouldBeLogged(priority)) {
 
-      log0(priority, findCallingMethod(), buildMessageText(args));
+      log0(priority, findCallingMethod(), buildAdditionalMessageText(args));
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/org/sosy_lab/common/log/TestLogManager.java
```diff
@@ -83,25 +82,46 @@ public void logf(Level pPriority, String pFormat, Object... pArgs) {
   }
 
   @Override
-  public void logUserException(Level pPriority, Throwable pE, @Nullable String pAdditionalMessage) {
+  public void logUserException(Level pPriority, Throwable pE, Object... pArgs) {
     checkNotNull(pPriority);
     checkNotNull(pE);
+    checkNotNull(pArgs);
```
PhilippWendler: This and similar methods should get the same checks as `log(Level, Object...)`, probably by refactoring them out into a separate method.

src/org/sosy_lab/common/log/TestLogManager.java
```diff
@@ -83,25 +82,46 @@ public void logf(Level pPriority, String pFormat, Object... pArgs) {
...
   }
 
   @Override
-  public void logDebugException(Throwable pE, @Nullable String pAdditionalMessage) {
+  public void logfUserException(Level pPriority, Throwable pE, String pFormat, Object... pArgs) {
+    checkNotNull(pPriority);
     checkNotNull(pE);
+    checkNotNull(pFormat);
+    checkNotNull(pArgs);
```
PhilippWendler: This and similar methods should get the same checks as `logf(Level, String, Object...)`, probably by refactoring them out into a separate method.

src/org/sosy_lab/common/log/LogManager.java
```diff
@@ -94,15 +93,42 @@
...
    * to file FOO.txt (REASON)".
    *
    * @param priority the log level for the message
    * @param e the occurred exception
-   * @param additionalMessage an optional message
+   * @param format The format string.
+   * @param args The arguments for the format string.
    */
-  void logUserException(Level priority, Throwable e, @Nullable String additionalMessage);
+  void logfUserException(Level priority, Throwable e, String format, Object... args);
```
PhilippWendler: Removing the old method breaks binary compatibility. We should therefore only add new methods, and remove none of the old ones.

src/org/sosy_lab/common/log/BasicLogManager.java
```diff
@@ -414,7 +418,95 @@ private void log0(Level priority, StackTraceElement stackElement, String msg) {
...
+      logDebugException(e, args);
+    } else {
+      // but keep calculated results
+      logDebugException(e, additionalMessage);
+    }
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public void logfUserException(Level priority, Throwable e, String format, Object... args) {
```
PhilippWendler: `logUserException` and `logfUserException` look almost identical, basically only one line changes if I see correctly. Can this be refactored?

src/org/sosy_lab/common/log/BasicLogManager.java
```diff
@@ -414,7 +418,95 @@ private void log0(Level priority, StackTraceElement stackElement, String msg) {
...
+    if (wouldBeLogged(priority)) {
+      additionalMessage = formatString(format, args);
+      StringBuilder logMessage = buildUserExceptionLogMessage(priority, e, additionalMessage);
+      @Var StackTraceElement frame = locateStackTraceElement(e);
+
+      log0(priority, frame, logMessage.toString());
+    }
+    // Enforce lazy string building for DEBUG level
+    if (additionalMessage == null) {
+      logDebugException(e, args);
```
PhilippWendler: Isn't `format` missing here? I am not sure this rather hard-to-understand logic is necessary. Maybe just have `wouldbeLogged(priority) || wouldbeLogged(EXCEPTION_DEBUG_LEVEL)` as condition for building the message?

PhilippWendler (changes requested):

src/org/sosy_lab/common/log/BasicLogManager.java
BCSol: Yes I missed the format there.
That is a better solution. 
I will change that

src/org/sosy_lab/common/log/BasicLogManager.java
BCSol: I shied away from doing this, cause of the awkard logic structure you mentioned, too.
With the mentioned changes, this will be easier.

src/org/sosy_lab/common/log/LogManager.java
BCSol: Would you prefer to extend `logUserException(Level priority, Throwable e, Object... args)` with the missing param.
´logUserException(Level priority, Throwable e, @Nullable String, Object... args)´
Or simply reintroduce the old method?

PhilippWendler: The old method needs to be added exactly as it was, otherwise it is not backwards compatible.

BCSol: Done.

src/org/sosy_lab/common/log/TestLogManager.java
BCSol: Also Done

src/org/sosy_lab/common/log/TestLogManager.java
BCSol: Also Done

src/org/sosy_lab/common/log/LogManager.java
```diff
@@ -123,9 +167,32 @@
...
+   *
+   * <p>The message is constructed lazily from <code>String.format(format, args)</code>. To make
+   * individual arguments lazy, use {@link MoreStrings#lazyString(Supplier)}.
+   *
    * @param e the occurred exception
+   * @param format The format string.
+   * @param args The arguments for the format string.
    */
-  void logDebugException(Throwable e);
+  void logfDebugException(Throwable e, String format, Object... args);
```
PhilippWendler: It seems the `logDebugException(Throwable)` override is still missing, right?

src/org/sosy_lab/common/log/LoggingOptions.java
```diff
@@ -92,6 +92,8 @@ public LoggingOptions(Configuration config) throws InvalidConfigurationException
     config.inject(this);
   }
 
+  public LoggingOptions() {}
```
PhilippWendler: This needs to be `protected`, not `public`. Probably should also get a comment that explains that it is for inheritance.

PhilippWendler (commented):
Could you please squash all commits that extend `LogManager` into one such that we have a nice history? (But keep the different change to `LoggingOptions` separately.)

src/org/sosy_lab/common/log/LogManager.java
PhilippWendler: This is still unsolved.

PhilippWendler (approved):

PR #15, opened 2017-12-10
Initial OrderStatisticMap/Set

lembergerth:
Provides the `OrderStatisticMap` and `OrderStatisticSet` interface
with naive implementations.

The interfaces provide creator methods that return objects of the implementing `Naive*` classes.
The `OrderStatistic*TestSuite` classes aim to provide a small test suite for the current and future `OrderStatistic*` implementations.

### DIFF
--- a/src/org/sosy_lab/common/collect/NaiveOrderStatisticMap.java
+++ b/src/org/sosy_lab/common/collect/NaiveOrderStatisticMap.java
@@ -35,8 +35,8 @@ import java.util.TreeMap;
  *
  * <p>The class wraps a {@link NavigableMap} object and delegates all methods inherited from the
  * <code>NavigableMap</code> interface to that. For the methods particular to the <code>
- * OrderStatisticMap</code> interface, it provides naive implementations that do not guarantee any
- * specific performance.
+ * OrderStatisticMap</code> interface, it provides naive implementations that do guarantee
+ * performance only in O(n).
  *
  * @param <K> type of the keys of this map. See the Javadoc of {@link OrderStatisticMap} for
  *     possible constraints on this type
--- a/src/org/sosy_lab/common/collect/NaiveOrderStatisticMap.java
+++ b/src/org/sosy_lab/common/collect/NaiveOrderStatisticMap.java
@@ -1,0 +1,162 @@
+/*
+ *  SoSy-Lab Common is a library of useful utilities.
+ *  This file is part of SoSy-Lab Common.
+ *
+ *  Copyright (C) 2007-2017  Dirk Beyer
+ *  All rights reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.sosy_lab.common.collect;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.collect.ForwardingNavigableMap;
+import com.google.common.collect.Maps;
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import java.io.Serializable;
+import java.util.Comparator;
+import java.util.Map;
+import java.util.NavigableMap;
+import java.util.TreeMap;
+
+/**
+ * An {@link OrderStatisticMap} with naive implementations of its functions.
+ *
+ * <p>The class wraps a {@link NavigableMap} object and delegates all methods inherited from the
+ * <code>NavigableMap</code> interface to that. For the methods particular to the <code>
+ * OrderStatisticMap</code> interface, it provides naive implementations that do not guarantee any
+ * specific performance.
+ *
+ * @param <K> type of the keys of this map. See the Javadoc of {@link OrderStatisticMap} for
+ *     possible constraints on this type
+ * @param <V> type of the values of this map
+ * @see OrderStatisticMap
+ */
+final class NaiveOrderStatisticMap<K, V> extends ForwardingNavigableMap<K, V>
+    implements OrderStatisticMap<K, V>, Serializable {
+
+  private static final long serialVersionUID = -3542217590830996599L;
+
+  private final NavigableMap<K, V> delegate;
+
+  private NaiveOrderStatisticMap(NavigableMap<K, ? extends V> pNavigableMap) {
+    delegate = new TreeMap<>(pNavigableMap);
+  }
+
+  /** Creates a new empty OrderStatisticMap using natural ordering. */
+  static <K, V> NaiveOrderStatisticMap<K, V> createMap() {
+    return new NaiveOrderStatisticMap<>(new TreeMap<>());
+  }
+
+  /** Creates a new empty OrderStatisticMap using the given comparator over its keys. */
+  static <K, V> NaiveOrderStatisticMap<K, V> createMap(Comparator<? super K> pComparator) {
+    return new NaiveOrderStatisticMap<>(new TreeMap<>(checkNotNull(pComparator)));
+  }
+
+  /**
+   * Creates a new OrderStatisticSet containing the same entries as the given map, using natural
+   * ordering over its keys.
+   */
+  static <K, V> NaiveOrderStatisticMap<K, V> createMapWithNaturalOrder(
+      Map<? extends K, ? extends V> pMap) {
+    return new NaiveOrderStatisticMap<>(new TreeMap<>(checkNotNull(pMap)));
+  }
+
+  /**
+   * Creates a new OrderStatisticMap containing the sames entries as the given map and using the
+   * same ordering over its keys as the given map.
+   */
+  static <K, V> NaiveOrderStatisticMap<K, V> createMapWithSameOrder(
+      NavigableMap<K, ? extends V> pNavigableMap) {
+    return new NaiveOrderStatisticMap<>(checkNotNull(pNavigableMap));
+  }
+
+  @Override
+  protected NavigableMap<K, V> delegate() {
+    return delegate;
+  }
+
+  @Override
+  public K getKeyByRank(int pIndex) {
+    return NaiveOrderStatisticSet.createSetWithSameOrder(delegate.navigableKeySet())
+        .getByRank(pIndex);
+  }
+
+  @Override
+  public Entry<K, V> getEntryByRank(int pIndex) {
+    K key = getKeyByRank(pIndex);
+    return Maps.immutableEntry(key, get(key));
+  }
+
+  @Override
+  @CanIgnoreReturnValue
+  public K removeByRank(int pIndex) {
+    K key = getKeyByRank(pIndex);
+    V val = remove(key);
+    assert val != null : "Key could be retrieved by rank, but no (or null) value associated";
+    return key;
+  }
+
+  @Override
+  public int rankOf(K pObj) {
+    checkNotNull(pObj);
+    return NaiveOrderStatisticSet.createSetWithSameOrder(delegate.navigableKeySet()).rankOf(pObj);
+  }
+
+  @Override
+  public OrderStatisticSet<K> navigableKeySet() {
+    return NaiveOrderStatisticSet.createSetWithSameOrder(super.navigableKeySet());
+  }
+
+  @Override
+  public OrderStatisticSet<K> descendingKeySet() {
+    return NaiveOrderStatisticSet.createSetWithSameOrder(super.descendingKeySet());
+  }
+
+  @Override
+  public OrderStatisticMap<K, V> descendingMap() {
+    return new NaiveOrderStatisticMap<>(super.descendingMap());
+  }
+
+  @Override
+  public OrderStatisticMap<K, V> subMap(
+      K fromKey, boolean fromInclusive, K toKey, boolean toInclusive) {
+    return new NaiveOrderStatisticMap<>(super.subMap(fromKey, fromInclusive, toKey, toInclusive));
+  }
+
+  @Override
+  public OrderStatisticMap<K, V> headMap(K toKey, boolean inclusive) {
+    return new NaiveOrderStatisticMap<>(super.headMap(toKey, inclusive));
+  }
+
+  @Override
+  public OrderStatisticMap<K, V> tailMap(K fromKey, boolean inclusive) {
+    return new NaiveOrderStatisticMap<>(super.tailMap(fromKey, inclusive));
+  }
+
+  @Override
+  public OrderStatisticMap<K, V> headMap(K toKey) {
+    return headMap(toKey, /* inclusive= */ false);
+  }
+
+  @Override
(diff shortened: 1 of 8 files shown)


src/org/sosy_lab/common/collect/OrderStatisticMap.java
```diff
@@ -0,0 +1,138 @@
...
+ *  limitations under the License.
+ */
+package org.sosy_lab.common.collect;
+
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import java.util.Comparator;
+import java.util.Map;
+import java.util.NavigableMap;
+
+public interface OrderStatisticMap<K, V> extends NavigableMap<K, V> {
```
PhilippWendler: Some JavaDoc here that explains what this is would be nice.

src/org/sosy_lab/common/collect/OrderStatisticMap.java
```diff
@@ -0,0 +1,138 @@
...
+  /**
+   * Returns the element of this set with the given rank. The lowest element in the set has rank ==
+   * 0, the largest element in the set has rank == size - 1.
+   *
+   * @param pIndex the rank of the element to return
+   * @return the element of this set with the given rank
+   * @throws IndexOutOfBoundsException if the given rank is out of the range of this set (i.e.,
+   *     pRank &lt; 0 || pRank &gt;= size)
+   */
+  K getKeyByRank(int pIndex);
```
PhilippWendler: Is this operation common enough to have its own method?
If yes, it could be a default method delegating to `getEntryByRank`, otherwise we could leave it out.

src/org/sosy_lab/common/collect/NaiveOrderStatisticMap.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.Map;
+import java.util.NavigableMap;
+import java.util.TreeMap;
+
+/**
+ * An {@link OrderStatisticMap} with naive implementations of its functions.
+ *
+ * <p>The class wraps a {@link NavigableMap} object and delegates all methods inherited from the
+ * <code>NavigableMap</code> interface to that. For the methods particular to the <code>
+ * OrderStatisticMap</code> interface, it provides naive implementations that do not guarantee any
```
PhilippWendler: Can we do better than "no guarantee"? At least something like O(n) should be possible, right? This should then also be stated on the public `create` methods of the interface.

src/org/sosy_lab/common/collect/NaiveOrderStatisticMap.java
```diff
@@ -0,0 +1,162 @@
...
+  private static final long serialVersionUID = -3542217590830996599L;
+
+  private final NavigableMap<K, V> delegate;
+
+  private NaiveOrderStatisticMap(NavigableMap<K, ? extends V> pNavigableMap) {
+    delegate = new TreeMap<>(pNavigableMap);
+  }
+
+  /** Creates a new empty OrderStatisticMap using natural ordering. */
+  static <K, V> NaiveOrderStatisticMap<K, V> createMap() {
```
PhilippWendler: These non-public factory methods are not strictly necessary, the constructor could be made package-private. But if these factory methods help to reduce code, then keep them.

src/org/sosy_lab/common/collect/NaiveOrderStatisticMap.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  @Override
+  protected NavigableMap<K, V> delegate() {
+    return delegate;
+  }
+
+  @Override
+  public K getKeyByRank(int pIndex) {
+    return NaiveOrderStatisticSet.createSetWithSameOrder(delegate.navigableKeySet())
```
PhilippWendler: Doesn't this copy the whole set on every access? This seems like a really expensive operation, and most callers of getters will assume that they do not involve large memory allocations. Is there no easy way to avoid this?

PhilippWendler (changes requested):

src/org/sosy_lab/common/collect/NaiveOrderStatisticMap.java
lembergerth: I'm not sure whether it's really *less* code, but it circumvents the existence of
`OrderStatisticMap(Map)` and `OrderStatisticMap(NavigableMap)`, and thus is more concrete and less error-prone.

src/org/sosy_lab/common/collect/OrderStatisticMap.java
lembergerth: I wanted to provide it in reference to `NavigableMap`.
It also includes methods `lower*`, `floor*`, etc. for `Entry` and `Key`.
But currently I'm not using or needing it.

src/org/sosy_lab/common/collect/OrderStatisticMapTestSuite.java
```diff
@@ -0,0 +1,369 @@
...
+  private static <K, V> void removeEntry(Map<K, V> pMap, Entry<K, V> pEntry) {
+    Assert.assertEquals(pMap.get(pEntry.getKey()), pMap.remove(pEntry.getKey()));
+  }
+
+  @Test
+  public void testEquals() {
+    OrderStatisticMap<String, String> l1 = createMap();
+    @Var OrderStatisticMap<String, String> l2 = createMap();
+
+    // Check that the map sorts its elements
```
PhilippWendler: `equals` of maps does not depend on ordering. I recommend `EqualsTester` for testing `equals`, because one should also check for example that such a map is equal to a `HashMap` with the same contents etc. But tests that the `NavigableMapTestSuiteBuilder` already does of course need not be written manually.

src/org/sosy_lab/common/collect/OrderStatisticMapTestSuite.java
```diff
@@ -0,0 +1,369 @@
...
+  @Test
+  public void testSubmapView_mutation() {
+    NavigableMap<String, String> map = createMap(ELEMS);
+    NavigableMap<String, String> subMap =
+        map.subMap(ELEMS.get(1).getKey(), true, ELEMS.get(2).getKey(), true);
+
+    Entry<String, String> toAdd = ELEMS_BELOW.get(2);
+
+    putEntry(subMap, toAdd);
+    Assert.assertTrue(containsEntry(subMap, toAdd));
```
PhilippWendler: `assertTrue` and `assertFalse` create hard-to-debug test cases and should be avoided. The various Truth methods are much better.

For consistency, it might make sense to switch to Truth for all assertions in these classes.

Furthermore, these methods can be statically imported.

src/org/sosy_lab/common/collect/OrderStatisticMapTestSuite.java
```diff
@@ -0,0 +1,369 @@
...
+    Assert.assertFalse(subMap.containsKey(toAdd.getKey()));
+
+    putEntry(map, toAdd);
+    Assert.assertTrue(containsEntry(subMap, toAdd));
+
+    removeEntry(map, toAdd);
+    Assert.assertFalse(containsEntry(subMap, toAdd));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
PhilippWendler: `expected` is bad except in single-statement test cases. Please use `try { ...; fail(); } catch (IllegalArgumentException expected) { }` with as little code inside the `try` as possible.

src/org/sosy_lab/common/collect/OrderStatisticSetTestSuite.java
```diff
@@ -0,0 +1,282 @@
...
+    Assert.assertEquals(ELEMS[1], subSet.first());
+    Assert.assertEquals(ELEMS[2], subSet.last());
+  }
+
+  @Test
+  public void testSubsetView_subsetOfSubset() {
+    OrderStatisticSet<String> set = createSet(ELEMS);
+    NavigableSet<String> subSet =
+        set.subSet(ELEMS[1], /* fromInclusive= */ true, ELEMS[3], /*
+    toInclusive=
```
PhilippWendler: please remove the line breaks in this comment

src/org/sosy_lab/common/collect/OrderStatisticMapTestSuite.java
```diff
@@ -0,0 +1,369 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.NavigableMap;
+import org.junit.Assert;
+import org.junit.Test;
+
+public abstract class OrderStatisticMapTestSuite {
+
+  @SuppressWarnings("unchecked")
```
PhilippWendler: This seems unnecessary, and all other occurrences in this file, too.

src/org/sosy_lab/common/collect/OrderStatisticMap.java
```diff
@@ -0,0 +1,173 @@
...
+   * operations specific to the OrderStatisticMap interface.
+   *
+   * @param pNavigableMap map to use entries and ordering of
+   * @param <K> type of the keys of the given and new map
+   * @param <V> type of the values of the given and new map
+   * @return a new OrderStatisticMap containing the same entries and using the same order over keys
+   *     as the given map
+   */
+  static <K, V> OrderStatisticMap<K, V> createWithSameOrder(
+      NavigableMap<K, ? extends V> pNavigableMap) {
```
PhilippWendler: This parameter could accept a `SortedMap`, right?
(same for `OrderStatisticSet`)

src/org/sosy_lab/common/collect/OrderStatisticSet.java
```diff
@@ -0,0 +1,150 @@
...
+  static <E> OrderStatisticSet<E> create(Comparator<? super E> pComparator) {
+    return NaiveOrderStatisticSet.createSet(pComparator);
+  }
+
+  /**
+   * Creates a new OrderStatisticSet containing the same elements as the given collection, using
+   * natural ordering. The returned map guarantees performance only in O(n) for the operations
+   * specific to the OrderStatisticSet interface.
+   */
+  static <E> OrderStatisticSet<E> createWithNaturalOrder(Collection<E> pCollection) {
```
PhilippWendler: We could accept an `Iterable` here. By using `Iterables.addAll`, we do not need more code and get the same performance benefit as before if a `TreeMap` is passed.

src/org/sosy_lab/common/collect/OrderStatisticSet.java
```diff
@@ -0,0 +1,150 @@
...
+   * @return the rank of the given element in the set, or -1 if the element is not in the set
+   * @throws NullPointerException if the given element is <code>null</code>
+   */
+  int rankOf(E pObj);
+
+  @Override
+  OrderStatisticSet<E> descendingSet();
+
+  @Override
+  OrderStatisticSet<E> subSet(
```
PhilippWendler, line 112: Btw., what is the rank of elements in a subset? Do the elements keep the same rank as in the original set, or do they get a new rank according to the contents of the subset? I would strongly prefer the latter. Are there tests for this?

The same question of course exists for all other subset views returned by some methods, and for the map methods. Furthermore, how does it work on the `descendingSet`?

The answers to these questions should be documented in the JavaDoc, either globally for all view-returning methods, or per method.

PhilippWendler (changes requested):
Please also run "ant format-source", it did a little bit of JavaDoc formatting.

src/org/sosy_lab/common/collect/OrderStatisticMapTestSuite.java
stieglma: why would you use try/fail instead of an expected excpeption rule? In my opinion this creates much nicer, and especially more readable code. Moreover the nested try-catches should be separate unit-tests, should they?

PhilippWendler: `ExpectedException` has a single advantage over `Test(expected=)`: it lets you reduce the scope where the exception is expected from the whole method to a range of code between the call to ExpectedException and the end of the method. So, it is better than `Test(expected=)`, and if the call to ExpectedException is the second-to-last statement, then it works fine.

However, both methods have one serious problem: If the expected exception is not thrown in the last statement of the method but before, the remainder of the test case will be silently ignored. This is error-prone, and it seems [Google has found out](https://github.com/google/error-prone/issues/219#issuecomment-71244204) that many developers assume that ExpectedException does some magic and actually manages to resume code execution after the expected exception is thrown. So my understanding from following the development of Guava, Truth, and Error-Prone is that both `ExpectedException` and `Test(expected=)` are [not](https://github.com/google/error-prone/issues/219#issuecomment-86110692) [worth](http://errorprone.info/bugpattern/TestExceptionChecker) it in most cases and the old `try-fail-catch` way is [recommended](https://stackoverflow.com/a/38866403/396730) (Christian Gruber is one of the Google developers of these projects) because it is less error-prone *and* allows you to safely write code after the expected exception. Error-Prone also has a check for not forgettting the `fail()` in the `try`, which would be the most important kind of error in this style of writing tests.

All this is true only until we have a newer JUnit that supports writing something like `assertThrows()` with a lambda. I guess then this should be preferred.

Now, independently from this it is of course good if tests are as small as possible, and this could also be improved in this class by splitting them into more methods. And if you follow that strictly, you might decide that the remaining risk of using `ExpectedException` is quite small and that it is fine to use. That would be fine.
For the current set of tests, however, I recommended using `try-fail-catch` because the exception handling is the more important problem than the somewhat large tests (the latter are only a problem for debugging but not necessarily error-prone), and because these tests already use `try-catch` in addition to `Test(expected=)`, switching completely to `try-fail-catch` would be easier.

stieglma: wow, thanks for the detailed response! I never thought about it that way. We never took the actual location where the exception is thrown into account (our unit-tests follow the arrange-act-assert principle and the the to-be-thrown exception is declared in the arrange block). What we do always is check the error message (and by "contract" if any method we implemented throws an exception, it has to have a proper message), and thus we still can be sure that we caught the right exception

src/org/sosy_lab/common/collect/OrderStatisticSet.java
lembergerth: I wouldn't even have thought of the first possibility.

`rankOf` and the other rank-methods work on the rank of the element in the `OrderStatistic*` collection the method is called on.
So if i call `subSet.rankOf(obj)`, it was obvious to me that, as the JavaDoc states, it returns the rank of `obj` in that specific set, i.e., `subSet`, not some backing set that the user may not even know of.

I have added clarifying Javadocs to the interface methods.

src/org/sosy_lab/common/collect/OrderStatisticMapTestSuite.java
```diff
@@ -145,43 +144,50 @@ public void testSubmapView_mutation() {
...
+            .put(ELEMS_BELOW.get(2))
+            .put(ELEMS.get(2))
+            .put(ELEMS_ABOVE.get(3))
+            .build();
     try {
       putEntry(subMap, ELEMS.get(0));
-    } catch (IllegalArgumentException e1) {
+      fail();
+    } catch (IllegalArgumentException expected) {
       try {
```
PhilippWendler: Nitpick: I wouldn't nest these, but just have consecutive `try-catch` statements.

PhilippWendler (approved):

src/org/sosy_lab/common/collect/OrderStatisticMapTestSuite.java
lembergerth: No, of course - this is a left-over of the `@Test(expected=..)` annotations

PR #21, opened 2018-04-29
Fix URL of comment in API webpage

lembergerth:


PhilippWendler:
Feel free do commit such changes directly.

PR #22, opened 2018-07-24
Timer provides #getMinTime() method

lembergerth:
To get the minimum measured time across all intervals.

It would be interesting for me to also get the minimum time that some repeated computation took,
but was this left out on purpose?

### DIFF
--- a/src/org/sosy_lab/common/time/Timer.java
+++ b/src/org/sosy_lab/common/time/Timer.java
@@ -192,8 +192,8 @@ public final class Timer {
   }
 
   /**
-   * Return the minimal time of all intervals. If the timer is running, the current interval is
-   * not considered. If the timer was never started, this method returns 0.
+   * Return the minimal time of all intervals. If the timer is running, the current interval is not
+   * considered. If the timer was never started, this method returns 0.
    */
   public TimeSpan getMinTime() {
     return export(minTime());
--- a/src/org/sosy_lab/common/time/Timer.java
+++ b/src/org/sosy_lab/common/time/Timer.java
@@ -29,8 +29,8 @@ import org.sosy_lab.common.time.Tickers.TickerWithUnit;
 
 /**
  * This class represents a timer like a stop watch. It can be started and stopped several times. It
- * measures return the sum, the average, the maximum and the number of those intervals. This class
- * is similar to {@link com.google.common.base.Stopwatch} but has more features.
+ * measures the sum, the average, the minimum, the maximum and the number of those intervals.
+ * This class is similar to {@link com.google.common.base.Stopwatch} but has more features.
  *
  * <p>This class is not thread-safe and may be used only from within a single thread.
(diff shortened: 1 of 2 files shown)


PhilippWendler:
The minimum was left out because it was not yet interesting so far, and keeping track of it costs something. What is your use case?

Does it make sense to count the current running interval for minimum time? E.g., if one starts the timer and immediately asks for minimum, it will return a low number that might not be near the actual minimum.

src/org/sosy_lab/common/time/Timer.java
```diff
@@ -183,10 +189,22 @@ public TimeSpan getMaxTime() {
...
+  public TimeSpan getMinTime() {
+    return export(minTime());
+  }
+
   long maxTime() {
     return Math.max(maxTime, currentInterval());
   }
 
+  long minTime() {
+    return Math.min(minTime, currentInterval());
```
PhilippWendler: This is broken. It always returns 0 except if the timer is running.

lembergerth:
> The minimum was left out because it was not yet interesting so far, and keeping track of it costs something. What is your use case?

I want to get a feeling for how fast two (mostly) independent computations can be.
While the maximum value is interesting, I'd also like to know whether it takes "at least 5 seconds",
or "at least 0.5 seconds".

> Does it make sense to count the current running interval for minimum time? E.g., if one starts the timer and immediately asks for minimum, it will return a low number that might not be near the actual minimum.

I don't have that use case, but just wanted to be consistent with max value.
If you don't think this is useful, I'll simply remove the corresponding methods.

PhilippWendler:
We can add it, but what should it return while the timer is running?

src/org/sosy_lab/common/time/Timer.java
```diff
@@ -85,6 +85,11 @@
   /** The maximal time of all intervals. */
   private long maxTime = 0;
 
+  /**
+   * The minimal time of all intervals.
+   */
+  private long minTime = 0;
```
PhilippWendler: This is also wrong: minTime will always stay zero.

PhilippWendler:
Oh, and what should the result be before the first start of the timer?

lembergerth:
In my opinion, a running timer should not be considered because it is not sure yet whether the value will be the minimum or not.
Can you think of a case where the user may want the timer to include a still running computation as the minimum? I imagine that that would be confusing and lead to wrong conclusions, but maybe I'm missing something...

Before the first start of the timer, I would return 0, because no time was consumed.
This would be similar to the maximum value of 0.

Both changes are incorporated into the last commit.
One could also keep `minTime = 0` and check in `stop()` whether the current intervall that is stopped
is the first interval, but I thought that the initial value of `MAX_VALUE` would be less error-prone.

PhilippWendler:
Ok.

Could you please run `ant format-source`?

PR #24, opened 2018-11-27
Make tests pass on OpenBSD

moritzbuhl:
JAVA_IO_TMPDIR.value() is "/tmp/" in OpenBSD. Adding another '/' to the end
would make the TEMP_DIR string in the FileTypeConverter class longer but
the path object will optimize the path.
Therefore "/tmp//.." is not "/tmp/.." (for a string comparison) and
"/tmp/../file" became "./file" in the FileTypeConverterTest class after a call
to CheckSafePath.

### DIFF
--- a/src/org/sosy_lab/common/configuration/converters/FileTypeConverter.java
+++ b/src/org/sosy_lab/common/configuration/converters/FileTypeConverter.java
@@ -69,7 +69,7 @@ public final class FileTypeConverter implements TypeConverter {
           File.class, Path.class, PathTemplate.class, PathCounterTemplate.class);
 
   private static final String TEMP_DIR =
-      StandardSystemProperty.JAVA_IO_TMPDIR.value() + File.separator;
+      Paths.get(StandardSystemProperty.JAVA_IO_TMPDIR.value() + File.separator).toString();
 
   @Option(secure = true, name = "output.path", description = "directory to put all output files in")
   private String outputDirectory = "output/";
--- a/src/org/sosy_lab/common/configuration/converters/FileTypeConverterTest.java
+++ b/src/org/sosy_lab/common/configuration/converters/FileTypeConverterTest.java
@@ -128,6 +128,8 @@ public class FileTypeConverterTest {
 
     @Parameters(name = "{0} (safe={1}, safeInFile={2})")
     public static List<Object[]> testPaths() {
+      String tmpDir =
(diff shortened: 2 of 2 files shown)


src/org/sosy_lab/common/configuration/converters/FileTypeConverter.java
```diff
@@ -69,7 +69,7 @@
           File.class, Path.class, PathTemplate.class, PathCounterTemplate.class);
 
   private static final String TEMP_DIR =
-      StandardSystemProperty.JAVA_IO_TMPDIR.value() + File.separator;
+      Paths.get(StandardSystemProperty.JAVA_IO_TMPDIR.value() + File.separator).toString();
```
PhilippWendler, line 72: This constant absolutely needs to have `File.separator` at its end, because we check for `path.startsWith(TEMP_DIR)`, which would wrongly accept paths outside the temp dir otherwise.

If a double slash is a problem here, it needs to be fixed somwhere else. But isn't this part of the change unnecessary anyway?

PhilippWendler (dismissed):

src/org/sosy_lab/common/configuration/converters/FileTypeConverter.java
moritzbuhl: In OpenBSD `StandardSystemProperty.JAVA_IO_TMPDIR.value()` evaluates to `"/tmp/"`. Then adding another `"/"` to it will result in a String `"/tmp//"`. Making it a Path object (and then using `toString()`) will remove duplicate `"/"` characters.
But I am not certain if on other systems the value from `Path.toString()` always ends with the separator.

PhilippWendler:
Thanks for the explanation, I understand now better. This is indeed a bug and needs to be fixed. Thanks for the PR. I added an alternative implementation to the one in this PR in d75a2c09aa087e60bf020b011029a266f3c19b46, which should behave the same but be more clear. Could you please test this?

moritzbuhl:
`ant tests` now passes on OpenBSD. Thanks.
```
[junitreport] Transform time: 10135ms
     [exec] 11952 / 21578 instructions covered
     [exec] 55.3897% covered
     [move] Moving 1 file to java-common-lib

tests:

BUILD SUCCESSFUL
Total time: 3 minutes 30 seconds
```

PR #30, opened 2020-03-25
Add MultiLineLogFormatter

BCSol:
MultiLineLogFormatter that supports color mode when attached to console.
Prints timestamp, log level, first log message line, and component
description in this order as first line. Appends remaining log message
lines as additional lines.

Make the LogFormatter used in the VerifierCloud available for the witness-browser.

### DIFF
--- a/src/org/sosy_lab/common/log/AbstractColoredLogFormatter.java
+++ b/src/org/sosy_lab/common/log/AbstractColoredLogFormatter.java
@@ -20,7 +20,7 @@ import java.util.logging.LogRecord;
  */
 abstract class AbstractColoredLogFormatter extends Formatter {
 
-  private boolean useColors;
+  private final boolean useColors;
 
   protected AbstractColoredLogFormatter(@Var boolean pUseColors) {
     if (pUseColors) {
--- a/src/org/sosy_lab/common/log/AbstractColoredLogFormatter.java
+++ b/src/org/sosy_lab/common/log/AbstractColoredLogFormatter.java
@@ -1,0 +1,77 @@
+// This file is part of SoSy-Lab Common,
+// a library of useful utilities:
+// https://github.com/sosy-lab/java-common-lib
+//
+// SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package org.sosy_lab.common.log;
+
+import com.google.common.base.MoreObjects;
+import com.google.errorprone.annotations.Var;
+import java.util.logging.Formatter;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+
+/**
+ * Abstract class for creating {@link Formatter}s that color {@link LogRecord}s with {@link
+ * Level#SEVERE} and {@link Level#WARNING} red.
+ */
+abstract class AbstractColoredLogFormatter extends Formatter {
+
+  private boolean useColors;
+
+  protected AbstractColoredLogFormatter(@Var boolean pUseColors) {
+    if (pUseColors) {
+      // Using colors is only good if stderr is connected to a terminal and not
+      // redirected into a file.
+      // AFAIK there is no way to determine this from Java, but at least there
+      // is a way to determine whether stdout is connected to a terminal.
+      // We assume that most users only redirect stderr if they also redirect
+      // stdout, so this should be ok.
+      if ((System.console() == null)
+          // Windows terminal does not support colors
+          || System.getProperty("os.name", "").startsWith("Windows")
+          // https://no-color.org/
+          || System.getenv("NO_COLOR") != null) {
+        pUseColors = false;
+      }
+    }
+    useColors = pUseColors;
+  }
+
+  @Override
+  public final String format(LogRecord lr) {
+    StringBuilder sb = new StringBuilder(200);
+
+    if (useColors) {
+      if (lr.getLevel().equals(Level.WARNING)) {
+        sb.append("\033[1m"); // bold normal color
+      } else if (lr.getLevel().equals(Level.SEVERE)) {
+        sb.append("\033[31;1m"); // bold red color
+      }
+    }
+    format(lr, sb);
+    if (useColors) {
+      sb.append("\033[m");
+    }
+    return sb.toString();
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this).add("useColors", useColors).toString();
+  }
+
+  /**
+   * Formats {@link LogRecord} using the provided {@link StringBuilder}.<br>
+   * This method corresponds to {@link Formatter#format(LogRecord)} in a template method pattern.
+   * The coloring behaviour is provided by superclass.
+   *
+   * @see Formatter#format(LogRecord)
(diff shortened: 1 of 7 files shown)


src/org/sosy_lab/common/log/MultiLineLogFormatter.java
```diff
@@ -0,0 +1,75 @@
+package org.sosy_lab.common.log;
```
PhilippWendler: License header is missing.

PhilippWendler (changes requested):
This duplicates too much code from `ConsoleLogFormatter`. Please refactor.

Furthermore, please add static factory methods.

src/org/sosy_lab/common/log/MultiLineLogFormatter.java
BCSol: - added licence header
- added static factory methods, but no `LoggingOption`s constructor
- Refactored coloured format feature into abstract class

We should maybe come up with a better name for the formatter. I named it `MultilineLogFormatter`, because the first line of the logmessage is displayed inline with:

`timestamp` `level` `message` [`component`:`class`.`method`]   
`additional optional message lines`

src/org/sosy_lab/common/log/AbstractColoredLogFormatter.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.errorprone.annotations.Var;
+import java.util.logging.Formatter;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+
+/**
+ * Abstract class for creating {@link Formatter}s that color {@link LogRecord}s with {@link
+ * Level#SEVERE} and {@link Level#WARNING} red.
+ */
+public abstract class AbstractColoredLogFormatter extends Formatter {
```
PhilippWendler: Please remove `public`.

src/org/sosy_lab/common/log/AbstractColoredLogFormatter.java
```diff
@@ -0,0 +1,79 @@
...
+          // https://no-color.org/
+          || System.getenv("NO_COLOR") != null) {
+        pUseColors = false;
+      }
+    }
+    useColors = pUseColors;
+  }
+
+  @Override
+  public String format(LogRecord lr) {
```
PhilippWendler: Should be `final` because subclasses are not supposed to override it, right?

src/org/sosy_lab/common/log/AbstractColoredLogFormatter.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+    return sb.toString();
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this).add("useColors", useColors).toString();
+  }
+
+  protected abstract void format(LogRecord lr, StringBuilder sb);
```
PhilippWendler, line 76: This should have JavaDoc explaining what subclasses are supposed to do.

PhilippWendler (changes requested):
Please update with `master` (or rebase) and adjust the license header with the current template. Sorry for the inconvenience.

src/org/sosy_lab/common/log/MultiLineLogFormatter.java
PhilippWendler: Hm, that does not really differ from the behavior of the other formatter, right? It also shows all lines of the message.

Is the main difference thus the timestamp? Then the name could be based on this.
Can other differences be eliminated?

BCSol: The main difference is in fact the timestamp, so this could be an option.

The other difference is that we wanted to have the most important information in chronological order on the first line:
- timestamp
- level: if the log message becomes very long, you have to look for the importance  
- message: only the first line, usually contains message, more lines can be seen as additional information, such as the stack trace
- component: since the vcloud doesn't fully leverage that feature yet and due to it's length it was moved to the end.

So the difference to the `FileLogFormatter` is only the position of the component descriptor, but it differs greatly to the `ConsoleLogFormatter`. We use this formatter for both the file and console logging. 
Technically we could refactor `FileLogFormatter` to allow the usage of colours as well and move to it. But then we should rename that one to TimestampedLogFormatter?

BCSol: We could generalize it completely, as well, so that users can provide a formatting string like:
`{timestamp:yyyy-MM-dd HH:mm:ss:SSS} {level} {message0} ({component}:{method})\n{message1+}`

src/org/sosy_lab/common/log/ConsoleLogFormatter.java
```diff
@@ -8,61 +8,32 @@
...
-      if ((System.console() == null)
-          // Windows terminal does not support colors
-          || System.getProperty("os.name", "").startsWith("Windows")
-          // https://no-color.org/
-          || System.getenv("NO_COLOR") != null) {
-        pUseColors = false;
-      }
-    }
-    useColors = pUseColors;
+  private ConsoleLogFormatter(@Var boolean useColors) {
```
PhilippWendler: Use `@Var` only if the variable is going to me modified (it is the inverse of `final`).

src/org/sosy_lab/common/log/TimestampedLogFormatter.java
```diff
@@ -0,0 +1,68 @@
...
+ * </pre>
+ */
+public class TimestampedLogFormatter extends AbstractColoredLogFormatter {
+
+  private static final DateTimeFormatter DATE_FORMAT =
+      DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss:SSS")
+          .withLocale(Locale.getDefault(Locale.Category.FORMAT))
+          .withZone(ZoneId.systemDefault());
+
+  private TimestampedLogFormatter(@Var boolean useColors) {
```
PhilippWendler: Same here.

src/org/sosy_lab/common/log/MultiLineLogFormatter.java
PhilippWendler: > Technically we could refactor `FileLogFormatter` to allow the usage of colours as well and move to it. But then we should rename that one to TimestampedLogFormatter?

That would be nice. The current `FileLogFormatter` would then stay, but be changed to extend `TimestampedLogFormatter` and marked as `@Deprecated`.

But I note that the current code of `TimestampedLogFormatter` is still different from `FileLogFormatter`. Do you intend to change that?

I would prefer to not invent our own customization mechanism for the message. We do not want to implement our own logging framework.

BCSol: >But I note that the current code of TimestampedLogFormatter is still different from FileLogFormatter. Do you intend to change that?

I guess the vcloud can live with the changed order of log entries and there might even be a incentive to you use the component logging mechanism more.

>That would be nice. The current FileLogFormatter would then stay, but be changed to extend TimestampedLogFormatter and marked as @Deprecated.

Sounds good, I'll make the necessary changes then.

PhilippWendler (approved):
I added a few small improvements. In the future, please push to a branch in this repository for creating a PR, this will give us a working CI for the PR. (Furthermore, creating a PR from master is discouraged.)

Issue #34, opened 2020-11-17
InvalidPathException from Configuration.parse on Windows


kfriedberger:
### Context
On Windows, paths typically start with a device identifier like `C:`.
This needs to be considered when handling paths in some places.

### Problem
Using a unzipped CPAchecker archive (r35978) the cmdline `.\scripts\cpa.bat -bmc E:\test\file.c`
throws an `InvalidPathException` when accessing files from the CPAchecker jar-file.

### Stacktrace
<pre>
java.nio.file.InvalidPathException: Illegal char <:> at index 3: jar:file:/C:/Users/kfrie/Downloads/CPAchecker-1.9.2-svn-35978M-windows/cpachecker.jar!/org/sosy_lab/cpachecker/util/automaton/find-target-locations.properties
        at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
        at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
        at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
        at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
        at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:229)
        at java.base/java.nio.file.Path.of(Path.java:147)
        at java.base/java.nio.file.Paths.get(Paths.java:69)
        at org.sosy_lab.common.configuration.Parser.parse(Parser.java:306)
        at org.sosy_lab.common.configuration.Parser.parse(Parser.java:160)
        at org.sosy_lab.common.configuration.ConfigurationBuilder.loadFromResource(ConfigurationBuilder.java:273)
        at org.sosy_lab.cpachecker.util.automaton.TargetLocationProviderImpl.tryGetAutomatonTargetLocations(TargetLocationProviderImpl.java:58)
[...]
</pre>

I do not see this problem as part of CPAchecker.
I hope that this problem can be solved in `java-common-lib`.
It appears for different OpenJDK versions (11 and 15) on Windows 10.

PhilippWendler:
Note that the exception message indicates that `C:` is *not* the problem, it is the occurrence of `:` in other places in the path.

[...]

PR #35, opened 2020-11-17
fix access to files that located inside a JAR.

kfriedberger:
This has failed on Windows 10 with OpenJDK 15 before.
See #34 for details.

### DIFF
--- a/src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
+++ b/src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
@@ -260,11 +260,9 @@ public final class ConfigurationBuilder {
 
     setupProperties();
 
-    // Get the path to the source, used for error messages and resolving relative path names.
     try {
-      // try to access the requested file.
       URI uri = url.toURI();
-      try (FileSystem fs = getFileSystemForUri(uri)) {
+      try (FileSystem fs = getFileSystemForUriInJars(uri)) {
         // Path uses FileSystemProvider internally to access the file, thus fs is unused.
         Path sourcePath = Paths.get(uri);
         parseSource(
@@ -299,21 +297,24 @@ public final class ConfigurationBuilder {
   }
 
   /**
-   * If the URI is part of a JAR file, we open the file system from the JAR. If this fails, we
-   * register/open a new file system for the JAR file. The paths for this file system are valid as
-   * long as the file system not closed.
+   * If the URI is part of a JAR file, we open the file system from the JAR. We only register/open a
+   * new file system for the JAR file, if it was not already open before. The paths for this file
+   * system are valid as long as the file system not closed. If anything fails, e.g., if the URI
+   * does not point to a JAR file or the JAR is already open, we return <code>null</code>.
    *
-   * @return the opened file system of the JAR file or <code>null</code>.
+   * A JAR-based file system uses ZipFileSystem that can be opened and closed several times.
+   *
+   * @return the opened file system of the JAR file if it was not open before or <code>null</code>.
    */
-  private static FileSystem getFileSystemForUri(URI uri) throws IOException {
+  private static FileSystem getFileSystemForUriInJars(URI uri) throws IOException {
     if ("jar".equals(uri.getScheme())) {
       for (FileSystemProvider provider : FileSystemProvider.installedProviders()) {
         if (provider.getScheme().equalsIgnoreCase("jar")) {
           try {
-            return provider.getFileSystem(uri);
-          } catch (FileSystemNotFoundException e) {
-            // register a new file system (provider) for the JAR.
+            // try to register a new file system (provider) for the JAR.
             return provider.newFileSystem(uri, ImmutableMap.of());
+          } catch (FileSystemNotFoundException e) {
+            // file system already exists: ignore it and return null after the loop
           }
         }
       }
--- a/src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
+++ b/src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
@@ -302,7 +302,7 @@ public final class ConfigurationBuilder {
    * system are valid as long as the file system not closed. If anything fails, e.g., if the URI
    * does not point to a JAR file or the JAR is already open, we return <code>null</code>.
    *
-   * A JAR-based file system uses ZipFileSystem that can be opened and closed several times.
+   * <p>A JAR-based file system uses ZipFileSystem that can be opened and closed several times.
    *
    * @return the opened file system of the JAR file if it was not open before or <code>null</code>.
    */
--- a/src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
+++ b/src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
@@ -24,6 +24,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystem;
+import java.nio.file.FileSystemAlreadyExistsException;
 import java.nio.file.FileSystemNotFoundException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -313,7 +314,7 @@ public final class ConfigurationBuilder {
           try {
             // try to register a new file system (provider) for the JAR.
             return provider.newFileSystem(uri, ImmutableMap.of());
-          } catch (FileSystemNotFoundException e) {
+          } catch (FileSystemAlreadyExistsException e) {
             // file system already exists: ignore it and return null after the loop
           }
         }
--- a/src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
+++ b/src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
@@ -26,6 +26,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystem;
 import java.nio.file.FileSystemAlreadyExistsException;
 import java.nio.file.FileSystemNotFoundException;
+import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.spi.FileSystemProvider;
@@ -75,7 +76,7 @@ public final class ConfigurationBuilder {
     setupProperties();
 
     properties.put(name, value);
-    sources.put(name, Paths.get(Configuration.NO_NAMED_SOURCE));
+    sources.remove(name);
 
     return this;
   }
@@ -97,9 +98,7 @@ public final class ConfigurationBuilder {
     setupProperties();
 
     properties.putAll(options);
-    for (String name : options.keySet()) {
-      sources.put(name, Paths.get(Configuration.NO_NAMED_SOURCE));
-    }
+    sources.keySet().removeAll(options.keySet());
 
     return this;
   }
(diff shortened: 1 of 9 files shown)


src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
```diff
@@ -260,9 +262,26 @@ public ConfigurationBuilder loadFromResource(Class<?> contextClass, String resou
...
+          if (provider.getScheme().equalsIgnoreCase("jar")) {
+            try {
+              provider.getFileSystem(uri);
+            } catch (FileSystemNotFoundException e) {
+              provider.newFileSystem(uri, ImmutableMap.of());
+            }
+          }
+        }
+      }
       sourcePath = Paths.get(url.toURI());
```
PhilippWendler:
```suggestion
      sourcePath = Paths.get(uri);
```

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
```diff
@@ -260,9 +262,26 @@ public ConfigurationBuilder loadFromResource(Class<?> contextClass, String resou
     @Var Path sourcePath;
     @Var String sourceString;
     try {
+      // try to access the requested file.
```
PhilippWendler: This does not try to access the given file, so please rewrite the comment.

Given that the whole code is quite non-obvious as it relies on the fact that `FileSystemProvider.newFileSystem` has some side effect, this really needs good documentation, with an explanation of the problem that this solves, why it solves it, and a link to the source of the JAR.

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
```diff
@@ -260,9 +262,26 @@ public ConfigurationBuilder loadFromResource(Class<?> contextClass, String resou
...
+      URI uri = url.toURI();
+      if ("jar".equals(uri.getScheme())) {
+        // if packed in a JAR file, we try to register the JAR as file system.
+        for (FileSystemProvider provider : FileSystemProvider.installedProviders()) {
+          if (provider.getScheme().equalsIgnoreCase("jar")) {
+            try {
+              provider.getFileSystem(uri);
```
PhilippWendler: Calling `getFileSystem` and throwing away its result is confusing, as this method should not have a side effect. Is the only reason why we call this method to determine whether the file system already exists?

In this case, I would find the following code much clearer:
```java
// Ensure file system for uri exists
try {
  provider.newFileSystem(uri, ImmutableMap.of());
} catch (FileSystemNotFoundException ignored) {
  // already exists
}
```

PhilippWendler (changes requested):
Did you test whether this actually solves #34? This is not obvious for me. After all, the exception from #34 is not the one thrown in the `try` block that is modified here, where a `Path` is created from a `URI`, but in a different method, where we try to construct the same `Path` from a string-

This solution has the side effect that a new file system is registered and stays in the list of file systems permanently, i.e., this is a (small) memory leak. Is there a way to avoid this?

Furthermore, this seems to open the referenced JAR file, although we will never read from it. Could this be avoided?

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
kfriedberger: I restructured the code and added some documentation.
Please ask if anything remains unclear.

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
```diff
@@ -250,35 +253,72 @@ public ConfigurationBuilder loadFromFile(Path file)
    * @throws IllegalArgumentException If the resource cannot be found or read, or contains invalid
    *     syntax or #include directives.
    */
+  @SuppressWarnings("try")
```
PhilippWendler, line 269: Why is this necessary?

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
PhilippWendler: Well, this comment is still as wrong as before, right?

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
```diff
@@ -250,35 +253,72 @@ public ConfigurationBuilder loadFromFile(Path file)
...
     // Get the path to the source, used for error messages and resolving relative path names.
-    @Var Path sourcePath;
-    @Var String sourceString;
     try {
-      sourcePath = Paths.get(url.toURI());
-      sourceString = sourcePath.toString();
-    } catch (URISyntaxException | FileSystemNotFoundException | IllegalArgumentException e) {
+      // try to access the requested file.
+      URI uri = url.toURI();
+      try (FileSystem fs = getFileSystemForUri(uri)) {
```
PhilippWendler: This closes the `FileSystem` afterwards, but the documentation of `FileSystemProvider.newFileSystem` says

> Once a file system is closed it is provider-dependent if the provider allows a new file system to be created with the same URI as a file system it previously created.

So this could mean that calling this method twice fails, right? Or is it guaranteed for the provider for JAR files that it will succeed to reopen the file twice? Is there any documentation on this?

Furthermore, if the file system already exists before this method is called, won't this even mean that we close it although somebody else has registered it?

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
```diff
@@ -250,35 +253,72 @@ public ConfigurationBuilder loadFromFile(Path file)
...
       sources.putAll(parser.getSources());
     } catch (InvalidConfigurationException | IOException e) {
       throw new IllegalArgumentException(
           "Error in resource " + resourceName + " relative to " + contextClass.getName(), e);
     }
+  }
 
-    return this;
+  /**
+   * If the URI is part of a JAR file, we open the file system from the JAR. If this fails, we
```
PhilippWendler: If this is just relevant for JARs, its name should reflect this.

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
```diff
@@ -250,35 +253,72 @@ public ConfigurationBuilder loadFromFile(Path file)
...
+      }
+    } catch (URISyntaxException
+        | FileSystemNotFoundException
+        | IllegalArgumentException
+        | IOException e) {
       // If this fails, e.g., because url is a HTTP URL, we can also use the raw string.
       // This will not allow resolving relative path names, but everything else works.
-      sourcePath = null;
-      sourceString = url.toString();
+      parseSource(contextClass, resourceName, source, Optional.empty(), url.toString());
```
PhilippWendler: Won't this still also the same kind of exceptions if we pass a string as source that was not able to be converted to a `Path` in this method?

PhilippWendler (dismissed):
You did not address my question about how this is related to solving #34, and also some other questions?

The added comments also does not explain the relevant part, because the exception in #34 is not from calling `Paths.get` inside `loadFromResource`. And there is no explanation why we need the file system to be open if `Path` uses that internally anyway and where this code come from.

Please either address or respond to all of my comments, please do not just ignore some of them.

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
kfriedberger: Opening a new filesystem registers a FileSystemProvider, thus the URI can be resolved.
This seems to be some background magic of Java.

I agree with your code change and will change it.

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
kfriedberger: Because the FileSystem in the try-with-ressources-block is never used locally.

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
kfriedberger: Yes, in general it depends on the FileSystems whether it can be reopened.
For JAR files, a ZipFileSystem is used and it seems to support multiple openings and closings.

Parallel access might be a problem: we could close a first instance and an interleaved concurrent reading process is then aborted. There is still not a solution for that.

I will change the code to not close, when already opened from other callee.

kfriedberger:
The implemented solution works so far, e.g., the cmdline from #34 works.
The exception reported in #34 is just an after-effect from the real problem,
thus, the modified try-block appears some lines before the actual problematic part,
e.g., where the first exception (`FileSystemNotFoundException`) was swallowed.
The changed code that works fine, because the file system for the JAR file is open before access.

src/org/sosy_lab/common/configuration/ConfigurationBuilder.java
PhilippWendler: This was not addressed? I see no way how this could be expected to work when `Paths.get()` is called on the result of `url.toString()` later, given that the conversion via `URI` has already failed?

kfriedberger: From my point, all cases that resulted in an execution of that branch, are handled as before, except the case, where a JAR file is accessed.
Thus, if it failed before, it will also fail now.
Currently I have no test case where I run into this branch.
Could you give an example why this case is required?

Do we have any JUnit test for this code?

PhilippWendler: Sure, the current code improves the situation, but now that we recognize that other cases will still fail for the same reason, we should fix these, too, right?

We cannot unit test this properly because it requires interfacing with Java APIs and use of URLs.

kfriedberger: Right, I thought the same: testing might not be possible.
Which further cases do you know that should be handled?
Can we write a documentation for manual testing at least?

PhilippWendler: Actually, the following code might work for testing:
```
    URL url =
        new URL(
            "https://raw.githubusercontent.com/sosy-lab/cpachecker/trunk/config/includes/resource-limits.properties");
    CharSource source = Resources.asCharSource(url, StandardCharsets.UTF_8);
    Configuration.builder().loadFromSource(source, url.toString(), url.toString());
```
This uses `loadFromSource`, but this in turn calls the same `parse` method as `loadFromResource`, it is just easier to specify the parameters. Note that the last parameter is `toString()` of an URL, just like in the `catch` block of `loadFromResource`.

For me, this creates a `UnixPath` with the `https://...`, which happens to work. But I guess it would crash on Windows?

kfriedberger: Yes, `https://...` crashes on Windows.
And I could not find any fix, because there is no matching FileSystemProvider available.
By default, `FileSystemProvider.installedProviders()` only lists the providers for `[UnixFileSystemProvider, JrtFileSystemProvider]` or `[WindowsFileSystemProvider, JrtFileSystemProvider]`, depending on the operating system. Any further filesystem access, e.g., accessing zip-files or http-files on Linux, might not be specified as default by Java itself, but provided by the environment.

Thus, I would not expect further support from Java and stay with the minimal working solution, which is already implemented in this PR.

PhilippWendler: Well, the fix could be to not rely on `Path` if it does not support what we need (as mentioned in https://github.com/sosy-lab/java-common-lib/issues/34#issuecomment-728745129).

This is somewhat tricky, because these `Path`s end up in public API, but at least we can avoid the crashes now and refactor some things to make all this more clear. I just pushed a commit that does this, please review (and ideally, test again on Windows).

PR #36, opened 2021-03-01
Replace PathCopyingPersistentTreeMap hash and size calculation

### LINKED ISSUE #64: Eager size tracking of PathCopyingPersistentTreeMap
`PathCopyingPersistentTreeMap` tracks sizes lazily, i.e., on first access to `size()` the whole tree is iterated and the result is cached for future accesses. We could improve this, with one of the following ways:

1. Track the subtree size in each node of the tree, similar to what is implemented in #51. This is guaranteed O(1), but increases the memory consumption of every node. This may or may not be relevant in practice, and it might also be that due to object sizes and alignment requirements there is actually no size increase, but we don't know.
2. Track the size only for each map instance, and when changing a map compute the new size without counting all nodes, similar to what is implemented in #36. This has the same memory usage as today and negligible overhead.

We should also consider `PartialSortedMap`, which we use for views like `subMap()`. There it is more difficult and we might only be able to optimize partially because a `PartialSortedMap` need to match a subtree of nodes exactly.

Druidos:
Replace PathCopyingPersistentTreeMap hash and size calculation from lazy linear algorithm to pre-calculations, which is efficient for public copying methods

### DIFF
--- a/src/org/sosy_lab/common/collect/PathCopyingPersistentTreeMap.java
+++ b/src/org/sosy_lab/common/collect/PathCopyingPersistentTreeMap.java
@@ -73,7 +73,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 public final class PathCopyingPersistentTreeMap<K extends Comparable<? super K>, V>
     extends AbstractImmutableSortedMap<K, V> implements PersistentSortedMap<K, V>, Serializable {
 
-  private static final long serialVersionUID = 1041711151457528188L;
+  private static final long serialVersionUID = -8878491978270416107L;
 
   @SuppressWarnings("unused")
   @SuppressFBWarnings(
@@ -86,11 +86,14 @@ public final class PathCopyingPersistentTreeMap<K extends Comparable<? super K>,
     private static final boolean RED = true;
     private static final boolean BLACK = false;
 
-    private static final long serialVersionUID = -7393505826652634501L;
+    private static final long serialVersionUID = 3222777055260841400L;
 
     private final @Nullable Node<K, V> left;
     private final @Nullable Node<K, V> right;
     private final boolean isRed;
+
+    private final int subTreeSize;
+    private final int subTreeHash;
 
     // Leaf node
     Node(K pKey, V pValue) {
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Thanks!

While this certainly increases the performance for some users, it causes higher memory usage for all users of this data structure due to the new fields in the node class. So I would like to understand the effects a little bit better before merging this.

Do you have any numbers on how much improves this performance or how much it increases memory usage in practice?
What is your use case that benefits from this? In CPAchecker, for example, we apply hash-code caching for these maps outside of them in those cases where we feel it is relevant, is this possible for you as well?

Druidos:
If memory consumption is critical, then I will move hash and size calculation from Node into PathCopyingPersistentTreeMap on put/remove-AndCopy.
Main idea is to optimize equals(). The hashcode calculation is obtained as a side result of the size calculation.
For example, following code twice iterates by whole pMap to identify size change:

```java
boolean compare(PathCopyingPersistentTreeMap<Integer, Integer> pMap) {
    PathCopyingPersistentTreeMap<Integer, Integer> comp = pMap.putAndCopy(1,1);
    return comp.equals(pMap);
}
```

Exact numbers about performance I can provide later.

PhilippWendler:
I now realized one thing: Caching the `hashCode()` strictly speaking violates the Java API for map's `hashCode()`, because the hash code of a map depends on the hash code of all keys and values, and in particular the values need not be immutable and may change. Then the hash code of a map needs to reflect this.

Now, it is unlikely that someone would use a map with mutable values, actually change them, and then also depend on the map's hash code. More likely are cases where either the map's hash code is not used at all, or where the hash code is used but everything is immutable. But it could happen, and it would certainly lead to very confusing errors and costly debugging. So it would be better to not do this.

Tracking the size differences in the way that is discussed here would be possible, though. If someone wants to pick this up they can do so.

PhilippWendler:
However, we need development to be done inside this repository such that we get CI results. So I am closing this PR, if work is taken up again, a new PR from an internal branch needs to be created.

PR #39, opened 2021-05-10
Suggestion: Add temporary directories

KlaraCimbalnik:
Suggestion: Add a directory equivalent to `TempFile` (and with that to `DeleteOnCloseFile`)

We could use this in CPAchecker [here](https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/47#note_550882390).

### DIFF
--- a/src/org/sosy_lab/common/io/TempFile.java
+++ b/src/org/sosy_lab/common/io/TempFile.java
@@ -196,6 +196,62 @@ public class TempFile {
     }
   }
 
+  /**
+   * Create a fresh temporary directory in JVM's temp directory.
+   *
+   * <p>The resulting {@link Path} object is wrapped in a {@link DeleteOnCloseDir}, which deletes
+   * the directory recursively including its contents as soon as {@link DeleteOnCloseDir#close()} is
+   * called.
+   *
+   * <p>It is recommended to use the following pattern: <code>
+   * try (DeleteOnCloseDir tempDir = TempFile.createDeleteOnCloseDir()) {
+   *   // use tempDir.toPath() to get the Path object denoting the temporary directory
+   * }
+   * </code>
+   *
+   * @param pPrefix The prefix of the randomly-generated directory name.
+   * @param pFileAttributes The {@link FileAttribute}s used for creating the directory.
+   */
+  public static DeleteOnCloseDir createDeleteOnCloseDir(
+      String pPrefix, FileAttribute<?>... pFileAttributes) throws IOException {
+    checkNotNull(pPrefix);
+    Path tempDir;
+    try {
+      tempDir = Files.createTempDirectory(TMPDIR, pPrefix, pFileAttributes);
+    } catch (IOException e) {
+      throw improveIoExceptionMessage(e, TMPDIR, pPrefix + "*");
+    }
+    return new DeleteOnCloseDir(tempDir);
+  }
+
+  /**
+   * A simple wrapper around {@link Path} that calls {@link
+   * com.google.common.io.MoreFiles#deleteRecursively(Path,
+   * com.google.common.io.RecursiveDeleteOption...)} recursively from {@link AutoCloseable#close()}
+   * to delete the directory including its contents.
+   */
+  @Immutable
+  public static final class DeleteOnCloseDir implements AutoCloseable {
+
+    private final Path path;
+
+    private DeleteOnCloseDir(Path pDir) {
+      path = pDir;
+    }
+
+    public Path toPath() {
+      return path;
+    }
+
+    /**
+     * Recursively delete all files and directories in the directory represented by this instance.
+     */
+    @Override
+    public void close() throws IOException {
+      com.google.common.io.MoreFiles.deleteRecursively(path);
+    }
+  }
+
   private static IOException improveIoExceptionMessage(
       IOException pException, Path pDir, String pPathString) {
     // The message of this exception is often quite unhelpful,
--- a/src/org/sosy_lab/common/io/TempFile.java
+++ b/src/org/sosy_lab/common/io/TempFile.java
@@ -109,18 +109,7 @@ public class TempFile {
       try {
         file = Files.createTempFile(dir, prefix, suffix, fileAttributes);
       } catch (IOException e) {
-        // The message of this exception is often quite unhelpful,
-        // improve it by adding the path were we attempted to write.
-        if (e.getMessage() != null && e.getMessage().contains(dir.toString())) {
-          throw e;
-        }
-
-        String fileName = dir.resolve(prefix + "*" + suffix).toString();
-        if (Strings.nullToEmpty(e.getMessage()).isEmpty()) {
-          throw new IOException(fileName, e);
-        } else {
-          throw new IOException(fileName + " (" + e.getMessage() + ")", e);
-        }
+        throw improveIoExceptionMessage(e, dir, prefix + "*" + suffix);
       }
 
       if (deleteOnJvmExit) {
@@ -151,7 +140,8 @@ public class TempFile {
      * the file as soon as {@link DeleteOnCloseFile#close()} is called.
      *
      * <p>It is recommended to use the following pattern: <code>
-     * try (DeleteOnCloseFile tempFile = Files.createTempFile(...)) {
+     * try (DeleteOnCloseFile tempFile =
+     *     TempFile.builder()[.. adjust builder ..].createDeleteOnClose()) {
      *   // use tempFile.toPath() for writing and reading of the temporary file
      * }
      * </code> The file can be opened and closed multiple times, potentially from different
@@ -203,6 +193,22 @@ public class TempFile {
     @SuppressWarnings("RedundantOverride") // to avoid deprecation warning when method is called
     public void close() throws IOException {
       super.close();
+    }
+  }
+
+  private static IOException improveIoExceptionMessage(
+      IOException pException, Path pDir, String pPathString) {
+    // The message of this exception is often quite unhelpful,
+    // improve it by adding the path were we attempted to write.
+    if (pException.getMessage() != null && pException.getMessage().contains(pDir.toString())) {
+      return pException;
+    }
+
+    String fileName = pDir.resolve(pPathString).toString();
+    if (Strings.nullToEmpty(pException.getMessage()).isEmpty()) {
+      return new IOException(fileName, pException);
+    } else {
+      return new IOException(fileName + " (" + pException.getMessage() + ")", pException);
     }
   }
 }


src/org/sosy_lab/common/io/TempDir.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Override
+    public void close() throws IOException {
+      deleteDirectory(path.toFile());
+    }
+  }
+
+  @CanIgnoreReturnValue
+  private static boolean deleteDirectory(File pDir) throws IOException {
```
PhilippWendler: This needs to be replaced with a call to `MoreFiles.deleteRecursively()` to avoid several problems.

src/org/sosy_lab/common/io/TempDir.java
```diff
@@ -0,0 +1,167 @@
...
+
+        String dirName = dir.resolve(prefix + "*").toString();
+        if (Strings.nullToEmpty(e.getMessage()).isEmpty()) {
+          throw new IOException(dirName, e);
+        } else {
+          throw new IOException(dirName + " (" + e.getMessage() + ")", e);
+        }
+      }
+
+      if (deleteOnJvmExit) {
```
PhilippWendler: Hm, do you really need this feature?

AFAIU the equivalent feature for files is somewhat fragile and should be avoided if there are other alternatives. For directories, it would be worse: First, the current implementation adds on thread for each directory, which can lead to a huge number of threads and overhead. Second, deleting a directory recursively can take an arbitrary long time and block the JVM shutdown, with a debugger needed for finding out why the program seems to hang. Third, if other threads continue to run while the shutdown hook is already active (this happens if the JVM receives Ctrl+C, for example), all kinds of problems can arise when two threads try to manipulate files in the same directory.

So I would say this feature should ideally not be added, only if there is a strong argument for it (and even then, maybe not add it in a general place where it can be easily misused, but specifically for where it is needed?).

src/org/sosy_lab/common/io/TempDir.java
```diff
@@ -0,0 +1,167 @@
...
+public class TempDir {
+
+  private TempDir() {}
+
+  /** Create a builder for temporary directories. */
+  public static TempDirBuilder builder() {
+    return new TempDirBuilder();
+  }
+
+  public static final class TempDirBuilder {
```
PhilippWendler: If the "delete on JVM exit" feature is removed, is it really worth having `TempDirBuilder`?

It would basically only wrap `new DeleteOnCloseDir(Files.createTempDirectory(dir, prefix, fileAttributes))`. This seems like a simple method would be enough?

(For temp files, the situation is different, because we have several more setters on the builder, so offering all feature combinations explods, and some setters have the same or overlapping parameter types, so there can easily be a confusion of parameters if added to a single method. But both problems do not exist here.)

src/org/sosy_lab/common/io/TempFile.java
```diff
@@ -30,7 +30,7 @@
 /** Utilities for temporary files. */
 public class TempFile {
 
-  private static final Path TMPDIR = Paths.get(StandardSystemProperty.JAVA_IO_TMPDIR.value());
+  public static final Path TMPDIR = Paths.get(StandardSystemProperty.JAVA_IO_TMPDIR.value());
```
PhilippWendler: Please don't make members public except if specifically desired (and then there should be some documentation). Here, package private would be enough.

src/org/sosy_lab/common/io/TempFile.java
```diff
@@ -151,7 +151,7 @@ public Path create() throws IOException {
      * the file as soon as {@link DeleteOnCloseFile#close()} is called.
      *
      * <p>It is recommended to use the following pattern: <code>
-     * try (DeleteOnCloseFile tempFile = Files.createTempFile(...)) {
+     * try (DeleteOnCloseFile tempFile = TempFile.builder()[.. adjust builder ..].createDeleteOnClose()) {
```
PhilippWendler: Thanks for finding and fixing this!

src/org/sosy_lab/common/io/TempDir.java
```diff
@@ -0,0 +1,167 @@
...
+     *
+     * <p>This instance can be safely used again afterwards.
+     */
+    public DeleteOnCloseDir createDeleteOnClose() throws IOException {
+      return new DeleteOnCloseDir(create());
+    }
+  }
+
+  /**
+   * A simple wrapper around {@link Path} that calls {@link Files#deleteIfExists(Path)} recursively
```
PhilippWendler: This method reference needs to be updated.

PhilippWendler (commented):
Thanks! I have some questions on how much of this is actually necessary.

src/org/sosy_lab/common/io/TempDir.java
KlaraCimbalnik: No, it is not needed

src/org/sosy_lab/common/io/TempDir.java
KlaraCimbalnik: Are you suggesting a single method like `TempDir.createDeleteOnClose()` or would you have multiple methods with a different number of parameters (e.g. so that it isn't necessary to always specify the directory where the temp dir should be created)?

PhilippWendler: Depends, this can be discussed depending on whether we think which parameters and parameter combinations are how relevant.

If I didn't overlook anything, three parameters are left: dir, prefix, and file attributes.

The file-attributes parameter is "free" from an API user's perspective because due to varargs one does not have to specify it. So there is no need for both `create(dir, prefix)` and `create(dir, prefix, fileAttributes)`, we can just always have the file attributes parameter (also `Files.createTempDirectory()` uses the same principle here, which confirms our choice).

The prefix parameter is something that we might always want to enforce. While technically not necessary, it is much nicer for users if a proper value is specified and they don't end up with completely randomly named temp directories. Again, `Files.createTempDirectory()` also always has this parameter, although they allow opt-out by passing `null`, but in general this confirms this choice again.

So this would leave at most `create(prefix, fileAttributes)` and `create(dir, prefix, fileAttributes)` as options. Always requiring to pass the directory should not be done because there is no obvious way to get the temporary directory in the correct platform-independent manner. Good API design makes good API uses easier than bad API uses, but requiring the directory parameter would make it easier for client code to misuse the API (by providing bad values like `.` or `/tmp`) than use it correctly (by providing `Paths.get(StandardSystemProperty.JAVA_IO_TMPDIR.value())`). So `create(prefix, fileAttributes)` is needed.

Do we actually need the directory parameter at all? I guess that the temp dir covers most of the cases of `Files.createTempDirectory()` anyway, and even more so for temp directories with automatic deletion as in our case. And because we already know that the other method variant is required anyway, there is no disadvantage in keeping `create(dir, prefix, fileAttributes)` out until we know a concrete use case.

So from this train of thought it seems that having one method with non-null prefix and file attributes as parameters seems like the best solution, but of course I am open for hearing other arguments.

(Note that nothing in this was meant to indicate that the method should be named `create`, this was just a place holder. I might even think that having two classes is not worth it here, we could add the static method to `DeleteOnCloseDir` itself, to `TempFile`, or to `IO` maybe.)

KlaraCimbalnik: Your argumentation makes sense to me, thank you for the detailed explanation.

I moved `DeleteOnCloseDir` to `TempFile` and added the static method `createDeleteOnCloseDir` with non-null prefix and file attributes as parameters to `TempFile`. Creating a temporary directory with `TempFile.createDeleteOnCloseDir(...)` seems natural to me as I would use the directory for storing temporary files.
Do you agree with that?

Following the move I also extracted the common code which improves the message of the `IOException`s.

src/org/sosy_lab/common/io/TempFile.java
```diff
@@ -9,6 +9,7 @@
 package org.sosy_lab.common.io;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.io.MoreFiles.deleteRecursively;
```
PhilippWendler: Static imports are useful for small and common methods like `checkNotNull`, but I would suggest to use them only when it actually makes the code better. For the single call to `MoreFiles.deleteRecursively()` it seems that without the import the code would be just as readable?

src/org/sosy_lab/common/io/TempFile.java
```diff
@@ -205,4 +196,73 @@ public void close() throws IOException {
...
+   *
+   * @param pPrefix The prefix of the randomly-generated directory name.
+   * @param pFileAttributes The {@link FileAttribute}s used for creating the directory.
+   */
+  public static DeleteOnCloseDir createDeleteOnCloseDir(
+      String pPrefix, FileAttribute<?>... pFileAttributes) throws IOException {
+    checkNotNull(pPrefix);
+    Path tempDir;
+    try {
+      tempDir = Files.createTempDirectory(TMPDIR, pPrefix, pFileAttributes.clone());
```
PhilippWendler: Why the `clone()` here? Defensive array copies can make sense, but here we just use it once to pass it on.

src/org/sosy_lab/common/io/TempFile.java
```diff
@@ -205,4 +196,73 @@ public void close() throws IOException {
...
+    private DeleteOnCloseDir(Path pDir) {
+      path = pDir;
+    }
+
+    public Path toPath() {
+      return path;
+    }
+
+    @Override
+    public void close() throws IOException {
```
PhilippWendler, line 250: Given that this deletes data, it might be good to have an explicit JavaDoc with "Recursively delete all files and directories in the directory represented by this instance."

PhilippWendler (commented):
Thanks! Just minor questions left now.

I am wondering what the best way to merge this PR would be. With the amount of back-and-forth, maybe a squash merge makes more sense than a regular merge. But this would mix the refactoring with the new feature.

Would it be too much effort for you to start from scratch, create one commit with the method refactoring, and one with the new code? Then just force-push on the same branch. If the effort is acceptable, this might be the nicest way. Otherwise we will simply merge as is (after the final polishing).

KlaraCimbalnik:
> Would it be too much effort for you to start from scratch, create one commit with the method refactoring, and one with the new code?

That's fine, I can do that. Just to be sure: When you are talking about the method refactoring, you are referring to the extraction of code to the new `improveIoExceptionMessage`-method, right?

PhilippWendler:
> When you are talking about the method refactoring, you are referring to the extraction of code to the new `improveIoExceptionMessage`-method, right?

Exactly, sorry for not being clear.

PhilippWendler (approved):
Thank you for polishing it so nicely!

PhilippWendler:
I published version 0.3000-486-g4e9577e of this library with this feature.

PR #40, opened 2021-06-24
ProcessExecutor decrease loglevel on timeout and interruption

BCSol:
Join re-throws a Interrupted/TimeoutException to the caller.
These cases may be considered by the caller as expected behaviour
and should not be logged as a WARNING.

### DIFF
--- a/src/org/sosy_lab/common/ProcessExecutor.java
+++ b/src/org/sosy_lab/common/ProcessExecutor.java
@@ -352,12 +352,12 @@ public class ProcessExecutor<E extends Exception> {
       return exitCode;
 
     } catch (TimeoutException e) {
-      logger.logf(Level.WARNING, "Killing %s[%d] due to timeout", name, pid);
+      logger.logf(Level.FINEST, "Killing %s[%d] due to timeout", name, pid);
       processFuture.cancel(true);
       throw e;
 
     } catch (InterruptedException e) {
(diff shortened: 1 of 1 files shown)


PhilippWendler:
Oh, log and rethrow is actually an anti-pattern. Thanks for opening this PR.

Typically we just remove the logging. However, in this case I am skeptical, because the info that processes will be cancelled (and why) can really be crucial during debugging (imagine finding out why processes die without this log message...). Furthermore, unfortunately not all code is perfect, and specifically `InterruptedException` probably has a relatively high change of being swallowed somewhere else.  So I would say, let's keep the logging line.

But:
- According to our [logging guide](https://gitlab.com/sosy-lab/software/cpachecker/-/blob/trunk/doc/Logging.md#log-levels) it is absolutely clear that the appropriate level is `FINEST`. It is a low-level utility class, but the message is not large enough to belong to `ALL`.
- Usually the best way to log such exceptions is `log*DebugException`. However, this would log on level `ALL` because of the large stack trace. Furthermore, it is not so important here to actually log the full exception details because this is not a case where the exception is swallowed, and this exception is unlikely to have any interesting information anyway, because we know that it will come from our executor service and not from somewhere else. We could use `log*UserException(Level.FINEST, ...)`, but this is uncommon and can be confusing.

So please change the log level to `FINEST`, then I will merge this.

PR #43, opened 2024-09-13
#38: improve architecture detection

### LINKED ISSUE #38: NativeLibraries could support more architectures, such as ARMv7 or ARMv8
The listing in [NativeLibraries](https://github.com/sosy-lab/java-common-lib/blob/4cf14d50647d69dc344dde026c0b39de8c1dce73/src/org/sosy_lab/common/NativeLibraries.java#L96) is inomplete and misses architectures like `ARMv7` (already outdated?) and `ARMv8` (identical to aarch64?).

For further info and testing, maybe @MartinSpiessl can help and provide access to our Raspi.

kfriedberger:
I am not sure how to test this code. 
We could mock System.getProperty with different data. However, static system-methods are complex to mock.

### DIFF
--- a/src/org/sosy_lab/common/NativeLibraries.java
+++ b/src/org/sosy_lab/common/NativeLibraries.java
@@ -9,6 +9,7 @@
 package org.sosy_lab.common;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
+import static com.google.common.base.Strings.nullToEmpty;
 
 import com.google.common.base.Ascii;
 import com.google.common.base.StandardSystemProperty;
@@ -16,6 +17,7 @@ import com.google.errorprone.annotations.Var;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Objects;
 import java.util.Optional;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
@@ -49,6 +51,10 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 @SuppressWarnings("NonFinalStaticField")
 public final class NativeLibraries {
 
+  private static final String REPORT_MESSAGE =
+      ", please report this together with information "
+          + "about your system (OS, architecture, JVM).";
+
   private NativeLibraries() {}
 
   @Deprecated // will become private
@@ -67,10 +73,7 @@ public final class NativeLibraries {
 
       @Var String prop = StandardSystemProperty.OS_NAME.value();
       if (isNullOrEmpty(prop)) {
-        throw new UnsatisfiedLinkError(
-            "No value for os.name, "
-                + "please report this together with information about your system "
-                + "(OS, architecture, JVM).");
+        throw new UnsatisfiedLinkError("No value for os.name" + REPORT_MESSAGE);
       }
 
       prop = Ascii.toLowerCase(prop.replace(" ", ""));
@@ -85,8 +88,8 @@ public final class NativeLibraries {
         throw new UnsatisfiedLinkError(
             "Unknown value for os.name: '"
                 + StandardSystemProperty.OS_NAME.value()
-                + "', please report this together with information about your system "
-                + "(OS, architecture, JVM).");
+                + "'"
+                + REPORT_MESSAGE);
       }
 
       return currentOS;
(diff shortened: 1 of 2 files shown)


src/org/sosy_lab/common/NativeLibraries.java
```diff
@@ -105,55 +111,62 @@ public static Architecture guessVmArchitecture() {
...
-                    + "', please report this together with information about your system "
-                    + "(OS, architecture, JVM).");
-          }
-        } else {
-          throw new UnsatisfiedLinkError("Could not detect system architecture");
+    /**
+     * In most cases, the system architecture matches the JVM architecture. In some rare cases,
+     * architecture from 'os.arch' might not be sufficient. For example, we could use an emulator
+     * and execute a 32bit JVM on a 64bit system. We cannot check each possible combination and
+     * emulator, but check only the important ones.
```
PhilippWendler: Hm, I think there is a misconception here. This comment seems to indicate that running a JVM that does not have exactly the same architecture as the OS is a special edge case that involves emulators.

But this is not the case! Running for example a 32-bit JVM on an 64-bit OS is a totally common case that is supported out-of-the box, no special setup or emulators involved.

What the code behind `guessVmArchitecture()` needs to do is to find out the *JVM* architecture, the OS architecture is not really relevant. The current code seems look at the latter, and then tries some adjustments for the JVM architecture. Can we do better, i.e., look at the JVM architecture primarily?

src/org/sosy_lab/common/NativeLibraries.java
```diff
@@ -105,55 +111,62 @@ public static Architecture guessVmArchitecture() {
...
+        case "i386":
+        case "i686":
+        case "x86":
+          return Architecture.X86;
+        case "amd64":
+        case "x86_64":
+          return Architecture.X86_64;
+        case "arm":
+        case "armv6":
+        case "armv7":
```
PhilippWendler: What are the differences between these? Is it safe to ignore them? Could this cause problems in the future? If we just return `ARM` now, we cannot change the later if we need to distinguish between them.

The questions to answer would be "Can `arm` binary libraries be loaded into a `armv7` JVM?" etc.

For `i386`, `i686`, and `x86` it is clear that these are almost synonyms and that the differences (which affect only processors from the nineties, so can be ignore. But for ARM I don't know.

Do you have a reference documentation for these values? Would be good to link it.

src/org/sosy_lab/common/NativeLibraries.java
```diff
@@ -95,8 +98,11 @@ public static OS guessOperatingSystem() {
 
   @Deprecated // will become private
   public enum Architecture {
+    // We only consider the mainly used architectures and ignore older ones like PPC or MIPS.
     X86,
-    X86_64;
+    X86_64,
+    ARM,
```
PhilippWendler: I guess this could be misunderstood as "any ARM architecture". Maybe we should use `ARM32`?

Or is it the case like for `x86` that everyone uses `ARM` to refer to 32-bit ARM? Then we would want to keep to the standard. What do compilers use here, for example?

src/org/sosy_lab/common/NativeLibraries.java
```diff
@@ -105,55 +111,62 @@ public static Architecture guessVmArchitecture() {
...
+      if (nullToEmpty(StandardSystemProperty.JAVA_VM_NAME.value()).matches("32-?bit")
+          || Objects.equals(System.getProperty("os.arch.data.model"), "32")
+          || Objects.equals(System.getProperty("sun.arch.data.model"), "32")) {
+        switch (arch) {
+          case X86_64:
+            return X86;
+          case ARM64:
+            return ARM;
+          default:
+            // $fall-through
```
PhilippWendler: There is no fall-through into another case here, just remove it.

src/org/sosy_lab/common/NativeLibraries.java
kfriedberger: The text contains the wording "for example ... emulator". Furthermore, executing 32bit application on a 64bit CPU requires the CPU to "redirect/simulate/emulate" the commands and registers, which is typically done quite performant directly in hardware.

Most developers install a JDK in 32 or 64bit according to their CPU architecture. Java in 32bit is outdated, as even embedded devices are available with more than 4 GB of memory, which is the maximum addressable space for a 32bit application. And the other way, from 32 to 64bit, does not work anyway.

Overall, I do not see a negative impact of my code changes for default machines and guessing architecture. For the ARM-part, we can discuss changes, see below.

src/org/sosy_lab/common/NativeLibraries.java
kfriedberger: Each ARM generation has its own architecture. Of course, there are differences between ARM 6, 7, 8, 9. 
There is one major architecture "AArch64" (ARM64) that is available on Armv8-A onwards, which was extended for ARM 9. I will remove older ARM versions from the code, as even ARM 7 is 20 years old and might no longer be available in devices. I do not have any matching hardware (except ARM8), so I can not provide other examples or testing.

When looking into GCC documentation, there axists a broad range of options and values (e.g., see https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html under option `-march`).
The binaries of Z3 are compiled with a plain option `arch=arm64` (https://github.com/Z3Prover/z3/blob/a3f35b6830fb9639a3e1d0dc23250e91973cb244/scripts/nightly.yaml#L124) and do not seem to include further architecture specifications. The Z3 build scripts also contain a simple check for architecture, quite similar to the code in this PR: https://github.com/Z3Prover/z3/blob/a3f35b6830fb9639a3e1d0dc23250e91973cb244/scripts/mk_unix_dist.py#L217-L224

src/org/sosy_lab/common/NativeLibraries.java
kfriedberger: I will remove plain `ARM`, as matching for ARM 6 or 7 is complex and might not be required by users.

src/org/sosy_lab/common/NativeLibraries.java
kfriedberger: done

src/org/sosy_lab/common/NativeLibraries.java
```diff
@@ -105,55 +110,53 @@ public static Architecture guessVmArchitecture() {
...
 
-      @Var String prop = System.getProperty("os.arch.data.model");
-      if (isNullOrEmpty(prop)) {
-        prop = System.getProperty("sun.arch.data.model");
+      currentArch = downgradeArchIfRequired(getArchFromSystemProperty());
+
+      return currentArch;
+    }
+
+    private static Architecture getArchFromSystemProperty() {
```
PhilippWendler:
```suggestion
    private static Architecture guessOsArchitecture() {
```
This needs to make explicit that the result is the OS architecture, and not the JVM one. While we are at it, there seems to be no reason to not make the naming consistent with the existing method.

src/org/sosy_lab/common/NativeLibraries.java
```diff
@@ -105,55 +110,53 @@ public static Architecture guessVmArchitecture() {
         return currentArch;
       }
 
-      @Var String prop = System.getProperty("os.arch.data.model");
-      if (isNullOrEmpty(prop)) {
-        prop = System.getProperty("sun.arch.data.model");
+      currentArch = downgradeArchIfRequired(getArchFromSystemProperty());
```
PhilippWendler: The `downgradeArchIfRequired()` here is misleading. We are not "downgrading" an architecture that we already have.

What we do here is that we guess the JVM architecture. And one of the inputs for this heuristic happens to be the OS architecture. So the logic for guessing the JVM architecture should be in a method whose name reflects that, and the OS architecture should be in a variable/parameter named `osArch`.

My suggestion is to move the logic for guessing the JVM architecture into `guessVmArchitecture()` and do `Architecture osArch = guessOsArchitecture();`. If you want to keep more parts of the logic in separate methods than everything in one method as it was (seems like a good idea), we could have a method `isVm32Bit()`. That should make the logic in `guessVmArchitecture()` pretty easy to follow.

src/org/sosy_lab/common/NativeLibraries.java
PhilippWendler: My point was that the comment gives a misleading representation of the importance of these code paths, which could easily lead to future developers ignoring this and introducing errors. Using a 32-bit JVM is sometimes done on purpose for saving memory if the application would not need more than a few GB anyway, so it is not like that 32-bit is exclusively a thing from the past.

The current comment is better.

src/org/sosy_lab/common/NativeLibraries.java
kfriedberger: This method does no longer guess, but simply use "os.arch". The new name seems to be a better match.

kfriedberger: I will rename it again.

src/org/sosy_lab/common/NativeLibraries.java
```diff
@@ -95,65 +98,67 @@ public static OS guessOperatingSystem() {
...
+              "Unknown value for os.arch: '"
+                  + StandardSystemProperty.OS_ARCH.value()
+                  + "'"
+                  + REPORT_MESSAGE);
+      }
+    }
+
+    /** Check whether the JVM is executing in 32 bit version. */
+    private static boolean isVm32Bit() {
+      return nullToEmpty(StandardSystemProperty.JAVA_VM_NAME.value()).matches(".*(32-?bit|i386).*")
```
PhilippWendler: This does not have the same semantics as the previous check because it case sensitive. Is this intended?

kfriedberger: thanks for the finding. fixed.

PhilippWendler (approved):

PhilippWendler:
Uhm, why did you squash this PR? I see no reason to do that and loose the history.

kfriedberger:
It was not a full squash, but I kept documentation separated from code changes. It might be cleaner to have only "important" commits and avoid intermediate commits like formatting or single-line fixes from PRs when only a few lines of code or a single file is touched. From other projects, I am used to cleanup the history before merging, then rebase and get a quite linear history for the main branch. I also did not find a guideline for this in the Readme.

PR #45, opened 2026-01-31
Add support for detecting BSD systems in NativeLibraries

daniel-raffler:
Hello,

this PR makes sure that the system load is tried if `NativeLibraries.loadLibrary()` fails because it can't recognize the operating system or the architecture of the system. Falling back to the system loader is the default behavior for `NativeLibraries.loadLibrary()`. However, the fallback will currently not work on unsupported systems, such as FreeBSD, as an exception is thrown instead

See https://github.com/sosy-lab/java-smt/issues/592 for the JavaSMT issue

### DIFF
--- a/src/org/sosy_lab/common/NativeLibraries.java
+++ b/src/org/sosy_lab/common/NativeLibraries.java
@@ -213,7 +213,7 @@ public final class NativeLibraries {
    * NativeLibraries}.
    *
    * @param libraryName A library name as for {@link System#loadLibrary(String)}.
-   * @return Found path or {@code Optional.absent()}
+   * @return Found path or {@link Optional#empty()}
    */
   @Deprecated // will become private
   public static Optional<Path> findPathForLibrary(String libraryName) {
--- a/src/org/sosy_lab/common/NativeLibraries.java
+++ b/src/org/sosy_lab/common/NativeLibraries.java
@@ -35,7 +35,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
  *       <p>Possible values for {@literal <arch>} include:
  *       <ul>
  *         <li>arm64
- *         <li>x86-64
+ *         <li>x86_64
  *         <li>x86
  *       </ul>
  *       <p>Possible values for {@literal <os>} include:
--- a/src/org/sosy_lab/common/NativeLibraries.java
+++ b/src/org/sosy_lab/common/NativeLibraries.java
@@ -219,13 +219,7 @@ public final class NativeLibraries {
   public static void loadLibrary(String name) {
     Optional<Path> path = findPathForLibrary(name);
     if (path.isPresent()) {
-      try {
-        System.load(path.orElseThrow().toAbsolutePath().toString());
-      } catch (UnsatisfiedLinkError e) {
-        // Revert to the system loader if we could not load the library that was found
-        System.loadLibrary(name);
-      }
-
+      System.load(path.orElseThrow().toAbsolutePath().toString());
     } else {
       System.loadLibrary(name);
     }
--- a/src/org/sosy_lab/common/NativeLibraries.java
+++ b/src/org/sosy_lab/common/NativeLibraries.java
@@ -29,8 +29,8 @@ import org.checkerframework.checker.nullness.qual.Nullable;
  *
  * <ul>
  *   <li>the "native library path" as returned by {@link #getNativeLibraryPath()}, which is the
- *       directory {@literal ../native/<arch>-<os>/} relative to the JAR file of this library,
- *       where {@literal <arch>} stands for your processor architecture and {@literal <os>} for the
+ *       directory {@literal ../native/<arch>-<os>/} relative to the JAR file of this library, where
+ *       {@literal <arch>} stands for your processor architecture and {@literal <os>} for the
  *       operating system.
  *       <p>Possible values for {@literal <arch>} include:
  *       <ul>
--- a/src/org/sosy_lab/common/NativeLibraries.java
+++ b/src/org/sosy_lab/common/NativeLibraries.java
@@ -217,15 +217,26 @@ public final class NativeLibraries {
    */
   @Deprecated // will become private
   public static Optional<Path> findPathForLibrary(String libraryName) {
+    @Var Path p;
+    try {
+      // Try to get the native library path
+      p = getNativeLibraryPath();
+    } catch (UnsatisfiedLinkError e) {
+      // Return Optional.empty if operating system or architecture are unknown to us
+      return Optional.empty();
+    }
     String osLibName = System.mapLibraryName(libraryName);
-    @Var Path p = getNativeLibraryPath().resolve(osLibName).toAbsolutePath();
+    // Look for the library in the native library path
+    p = p.resolve(osLibName).toAbsolutePath();
     if (Files.exists(p)) {
       return Optional.of(p);
     }
+    // Look for the library next to the *.jar file
     p = Classes.getCodeLocation(NativeLibraries.class).resolveSibling(osLibName).toAbsolutePath();
     if (Files.exists(p)) {
       return Optional.of(p);
     }
+    // If nothing was found, return Optional.empty
     return Optional.empty();
   }
 }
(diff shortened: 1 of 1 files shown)


PhilippWendler:
What is the motivation for this?

If we encounter a system that is currently unsupported, why not add support for it instead? With the fallback we would never notice.

daniel-raffler:
Hello Philipp,

they're using NetBSD, and we're unlikely to add support anytime soon as we have no way of compiling binaries for BSD. It would still be possible to extends `OS` and simply return a "native libary path" that points to an empty folder when using NetBSD. However, `findPathForLibrary` will then start looking in the folder for `common.jar` next. Only after this also fails will the system loader be tried

The issue with this is, that BSD and Linux both use `*.so` files for libraries. Because of that `findPathForLibrary` may return "successfully" on NetBSD after finding a Linux library. Loading the library should still fail, but the system loader is then never tried because we did find a valid path to the library initially

We could probably fix this by catching the exception in `loadLibrary()` and trying again with the system loader. Does this make more sense to you?

src/org/sosy_lab/common/NativeLibraries.java
```diff
@@ -28,18 +28,26 @@
...
  *   <li>the "native library path" as returned by {@link #getNativeLibraryPath()}, which is the
- *       directory {@literal ../native/<arch>-<os>/} relative to the JAR file of this library, with
- *       {@literal <arch>-<os>} being one of the following values depending on your system:
+ *       directory {@literal ../native/<arch>-<os>/} relative to the JAR file of this library,
+ *       where {@literal <arch>} stands for your processor architecture and {@literal <os>} for the
+ *       operating system.
+ *       <p>Possible values for {@literal <arch>} include:
+ *       <ul>
+ *         <li>arm64
+ *         <li>x86-64
```
kfriedberger: This might use an underscore.

src/org/sosy_lab/common/NativeLibraries.java
```diff
@@ -202,7 +219,13 @@ public static Path getNativeLibraryPath() {
   public static void loadLibrary(String name) {
     Optional<Path> path = findPathForLibrary(name);
     if (path.isPresent()) {
-      System.load(path.orElseThrow().toAbsolutePath().toString());
+      try {
+        System.load(path.orElseThrow().toAbsolutePath().toString());
+      } catch (UnsatisfiedLinkError e) {
+        // Revert to the system loader if we could not load the library that was found
+        System.loadLibrary(name);
```
kfriedberger: While I fully understand the wish and intention for this fallback, I see this as a critical change and would like to avoid it, because it makes it harder to detect broken dependencies or mis-configured or mis-compiled ones. If the path is present and can not be loaded, just inform the user and do not hide the issue.

src/org/sosy_lab/common/NativeLibraries.java
```diff
@@ -84,6 +95,12 @@ public static OS guessOperatingSystem() {
         currentOS = WINDOWS;
       } else if (prop.startsWith("macosx")) {
         currentOS = MACOSX;
+      } else if (prop.startsWith("freebsd")) {
+        currentOS = FREEBSD;
+      } else if (prop.startsWith("netbsd")) {
+        currentOS = NETBSD;
+      } else if (prop.startsWith("openbsd")) {
+        currentOS = OPENBSD;
```
kfriedberger, line 103: Do we have tests for that? Or some real-world one-man tester that can evaluate that it work as intended.

I like that improvement.

src/org/sosy_lab/common/NativeLibraries.java
daniel-raffler: Nice catch, thanks!

src/org/sosy_lab/common/NativeLibraries.java
daniel-raffler: I tried it on NetBSD in a VM and the system seems to be recognized

It's surprisingly difficult to find any official documentation for the OS names. However, there is an apache package [SystemUtils](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/SystemUtils.html) that has methods like `IS_OS_NET_BSD`. [Internally](https://github.com/apache/commons-lang/blob/4eb7bdfb8b6f54219342fbaf8a36f6af94e3e481/src/main/java/org/apache/commons/lang3/SystemUtils.java#L1674) it's using the same String constants for the comparison

src/org/sosy_lab/common/NativeLibraries.java
daniel-raffler: Hmm.. I can see your point

I've now removed the fallback, and instead added a check in `findPathForLibrary` to make sure that the path of the *.jar is only tried on Linux, Windows or MacOS. On BSD the function will return immediately when no binary can be found on the "native library path". This still means changing the API slightly, but I hope it's less of a problems as the affected systems (= BSD) didn't used to work at all

PhilippWendler:
> they're using NetBSD, and we're unlikely to add support anytime soon as we have no way of compiling binaries for BSD.

Caution: You need to take a step back and change your point of view. Instead of "we" in this sentence you need to say "JavaSMT". But this project here is not JavaSMT.

From the point of this project here it does not make sense to treat some OS specially. While we do add smaller OS only on an on-demand basis to `NativeLibraries.OS`, all added OS should have the same behavior and features. After all, we do not know which projects want to use `NativeLibraries` and on which OS their users are.

And this is also relevant for JavaSMT: Even if you do not provide *BSD binaries, you certainly want to have your users be able to compile the binaries and use them, for which you need `NativeLibraries` to behave like on other OS, right?

Or put differently: If you want to continue with this PR, you need to provide a clear argument why we should handle Linux/Windows/MacOS differently than *BSD in `NativeLibraries`.

IMO the process needs to be the following: In the first place, we add *BSD cases to `NativeLibraries.OS` and the respective detection logic. There is no question about this and such changes can be merged directly for any OS.

Then, if there are problems in practice with how native-library loading works in some scenarios, we need to discuss these scenarios on by one, understand the actual problem in them and identify solutions. For this an issue might be useful that clearly describes the problem.

To me the process of this so far seems backwards.

> The issue with this is, that BSD and Linux both use `*.so` files for libraries. Because of that `findPathForLibrary` may return "successfully" on NetBSD after finding a Linux library. Loading the library should still fail, but the system loader is then never tried because we did find a valid path to the library initially

Why should there be a Linux library on a NetBSD system?

> We could probably fix this by catching the exception in `loadLibrary()` and trying again with the system loader. Does this make more sense to you?

I agree with @kfriedberger in https://github.com/sosy-lab/java-common-lib/pull/45#discussion_r2760147767 on the topic of fallbacks.

daniel-raffler:
Hello Philipp,

I reverted the changes in `findPathForLibrary` so behavior should now be consistent across all platforms

> Why should there be a Linux library on a NetBSD system?

Because they're pulled by Ivy. In `CPAchecker`, for example, `common.jar` can be found under `cpachecker/lib/java/runtime`, and the same folder will also contain `libmathsat5j.so`, `libz3java.so`, and all other solver binaries. It is these Linux binary that will be found by the loader on *BSD

It's possible to work around this by using the [multiarch](https://github.com/sosy-lab/java-smt/blob/master/doc/Getting-started.md#architecture-specification-and-operating-system) configuration for JavaSMT. However, in that case the binaries will simply be moved to `cpachecker/lib/java/runtime/x64`, meaning the "same folder as the *.jar file" just doesn't apply any more

Loading the library from the directory of the `*.jar` file in my opinion only made sense as long as all platforms use a different file ending for libraries. Otherwise, there will be collisions, and it's just not possible to have all platforms in the same folder

> IMO the process needs to be the following: In the first place, we add *BSD cases to `NativeLibraries.OS` and the respective detection logic. There is no question about this and such changes can be merged directly for any OS.
> 
> Then, if there are problems in practice with how native-library loading works in some scenarios, we need to discuss these scenarios on by one, understand the actual problem in them and identify solutions. For this an issue might be useful that clearly describes the problem.

I'm fine with just merging the current changes. It worked for me on NetBSD when I tried it with Z3, and we can open another issue if there are still problems

PhilippWendler:
Then I would do a squash-merge of this MR (because it contains lots of changes that are irrelevant to the final state) once @kfriedberger agrees.

kfriedberger (approved):
_basst scho_ ... good for me

PR #47, opened 2026-03-10
Upgrade from Java 11 to Java 17

### LINKED ISSUE #46: Bump minimum Java version to 17
Now that [JavaSMT will require Java 17](https://github.com/sosy-lab/java-smt/pull/611), there is nothing that stops this project from requiring Java 17 as well and making use of [such nice features](https://gitlab.com/sosy-lab/software/cpachecker/-/merge_requests/100).

PhilippWendler:
JavaSMT is also doing this, so all projects depending on this library are now at Java 17 or newer.

The PR updates the build config and then applies several commits that make use of Java 17 features (pattern-matching instanceof, new switch syntax, `String.formatted()`, and `sealed` classes). Most of the commits are auto-generated.

For reviewing, the interesting questions are:
- Are the updates in da795f1729c8d6bbe19c7de1ffc50e23c0eb2f1d correct and complete, for example with regards to all IDE configs?
- Are the records introduced in da795f1729c8d6bbe19c7de1ffc50e23c0eb2f1d, 624df6b641d823d67ac211d64728cb8e5ac9c247, and aef9ced2eae956b653992893dbb80c70efa36852 backwards compatible with the previous classes? AFAIU and with my tests in CPAchecker this should be the case.
- Of course if there is anything that I have overlooked and that would be nice to improve with a Java 17 feature I am also glad to be pointed to it.

Fixes #46.


PhilippWendler:
@kfriedberger Are you interested in looking at this?

src/org/sosy_lab/common/configuration/converters/FileTypeConverter.java
```diff
@@ -184,9 +178,9 @@ private static InvalidConfigurationException forbidden(
       String reason, String optionName, Path path, Object... args)
       throws InvalidConfigurationException {
     throw new InvalidConfigurationException(
-        String.format(
-            "The option %s specifies the path '%s' that is forbidden in safe mode " + reason + ".",
-            FluentIterable.<Object>of(optionName, path).append(args).toArray(Object.class)));
+        ("The option %s specifies the path '%s' that is forbidden in safe mode " + reason + ".")
```
kfriedberger, line 182: Why do you mix formatter and concatenation here?

kfriedberger (approved):
I made a quick review. Changes look good to me.

src/org/sosy_lab/common/configuration/converters/FileTypeConverter.java
PhilippWendler: Ah, I also stumbled over this. The reason is that `reason` is a (part of the) format string. I added a comment explaining this.

Thanks for having a look!

PR #52, opened 2026-04-18
Draft: 48 extend persistent data structures with deque

### LINKED ISSUE #48: Extend Persistent Data-Structures with Deque
CPAchecker implements a `PersistentStack` based on the `PathCopyingPersistentTreeMap`. This should be moved into this repository, as others can benefit from it. But since a stack (and a queue) are just special views on a deque, we can just implement a `PersistentDeque` and get a queue and stack out of the deque, increasing our set of persistent data-structures in a thought out way.

The `PersistentDeque` can be based on the `PathCopyingPersistentTreeMap`, like in CPAchecker, or may be based on other starting points, for example the `PersistentLinkedList`, to optimize the implementation.

Much more optimal solution(s) can be found for example in [Implementing Persistent O(1) Stacks and Queues in R by  Shawn T. O'Neil](https://journal.r-project.org/articles/RJ-2015-009/index.html). Looking at the R code, we could try to translate the concept to Java with our `PersistentLinkedList` rather quickly i think.

c-m-elliott:
This provides a persistent implementation of a deque, a double-ended queue which supports insert and remove operations on either end and allows access of the top- and bottom-most objects. It uses a variable type and achieves greater efficiency by splitting its contents up into two separate PersistentLinkedLists so operations on either end will require O(1) in most cases. If one of the lists ends up empty while the other still contains two or more elements, a rebalancing operation is performed to split the non-empty list into two new parts. When this happens, it will increase runtime. In most cases, it shouldn't be necessary on each insert or delete operation.

Currently, PersistentDeque only supports creating an empty instance and adding new objects one by one. Further methods that enable creating instances that immediately contain one or more elements can be added in the future if desired.

It does not implement java.util.Deque as previously suggested, as this contained a number of methods that didn't seem essential to me and would have complicated matters significantly. This should also be possible to add at a later time if necessary.


.gitignore
```diff
@@ -37,3 +37,4 @@
 /lib/java
 /lib/java-contrib
 
+/.idea/
```
baierd: We don't want to forbid changes to the idea folder in general. Please revert this change. 

But i understand why you did this, we updated our default IntelliJ IDEA settings to reduce the automatic changes that you most likely experienced by a lot in CPAchecker. I'll take care of updating the IntelliJ settings in this project as well.

src/org/sosy_lab/common/collect/PersistentDequeInterface.java
```diff
@@ -0,0 +1,77 @@
...
+ *  ({@link #insertTop(Object)}, {@link #insertBottom(Object)}, {@link #deleteTop()},
+ *  {@link #deleteBottom()}) will
+ *  not modify the current
+ *  instance,
+ *  but return a new instance instead.
+ *
+ * @param <T> type of elements stored in deque
+ */
+@Immutable(containerOf = "T")
+public interface PersistentDequeInterface<T> {
```
baierd: You can just inherit the proper [Java interface](https://docs.oracle.com/javase/8/docs/api/java/util/Deque.html) instead of defining everything on your own.

It is also a good idea to split the ideas of immutability and persistence, as it allows us to build immutable and also persistent data-structures independently of each other, but also helps understanding what each interface does, by splitting the concerns.
This is why we have a [PersistentMap](https://github.com/sosy-lab/java-common-lib/blob/main/src/org/sosy_lab/common/collect/PersistentMap.java), and also a `AbstractImmutableMap` (more on that below). We then inherit from both to implement an immutable and persistent data-structure, for example our [PathCopyingPersistentTreeMap](https://github.com/sosy-lab/java-common-lib/blob/main/src/org/sosy_lab/common/collect/PathCopyingPersistentTreeMap.java).

Normally we would disallow certain methods due to immutability right away using a default implementation in an immutable interface (e.g. add and remove methods) . But we are still in Java 11 here (update to 17 is currently being worked on). So the solution is an abstract class instead. You can take a look [at our AbstractImmutableMap](https://github.com/sosy-lab/java-common-lib/blob/main/src/org/sosy_lab/common/collect/AbstractImmutableMap.java) for example. You also get a notion on how to use the annotations like `@DoNotCall`.

src/org/sosy_lab/common/collect/PersistentDeque.java
```diff
@@ -0,0 +1,190 @@
...
+    //If deque contains only one element, one of the lists will be empty. Calling head() on an
+    // empty list throws an exception, so this needs to be caught. Further, the one element in
+    // the non-empty list should be returned, as it is both head and tail of the deque.
+    try {
+      return top.head();
+    } catch(IllegalStateException e1) {
+      try {
+        return bottom.head();
+      } catch(IllegalStateException e2) {
+        return null;
```
baierd: Do you really want to return `null` instead of throwing an exception if something goes wrong? (in general)

Try to look at this from a user-perspective; you don't know about the implementation details of a data-structure that you use. You just expect the guarantees given in the documentation. Suddenly you get an unexpected `null` back (you did not say that `null` is a possible return value in your interface for `getTop()`). `null` is sometimes hard to debug, because if not checked for it, it causes exceptions only when used. But there is no guarantee that the returned value is used right away. It might first be stored in some other data-structure etc. and used much later, making debugging much harder!

You can take a look at how we handle `null` values (we don't allow them in the first place ;D) here, and Javas Deque implementation shows you how to handle requesting a value for an empty Deque (or data-structure in general). 

If an exception is thrown by a used method, it makes sense to just declare the using method as also throwing as long as you can not guarantee that your implementation does never throw this exception. Your documentation should be extended by appropriate reasoning if you also throw the exception (i.e. look at why and when the used method throws)

What might have added to the confusion is that our `PersistentLinkedList` just throws the wrong exception here ;D 

The same applies to `getBottom()` below.

src/org/sosy_lab/common/collect/PersistentDeque.java
```diff
@@ -0,0 +1,190 @@
...
+  @Override
+  public T getTop() {
+    //If deque contains only one element, one of the lists will be empty. Calling head() on an
+    // empty list throws an exception, so this needs to be caught. Further, the one element in
+    // the non-empty list should be returned, as it is both head and tail of the deque.
+    try {
+      return top.head();
+    } catch(IllegalStateException e1) {
+      try {
+        return bottom.head();
```
baierd: Are you sure you want the head here in all cases?

src/org/sosy_lab/common/collect/PersistentDeque.java
```diff
@@ -0,0 +1,190 @@
...
+   *
+   * @return element at top of deque; null if deque empty
+   */
+  @Nullable
+  @Override
+  public T getTop() {
+    //If deque contains only one element, one of the lists will be empty. Calling head() on an
+    // empty list throws an exception, so this needs to be caught. Further, the one element in
+    // the non-empty list should be returned, as it is both head and tail of the deque.
+    try {
```
baierd: Your two try-catch blocks could be simplified by using only a single one.

src/org/sosy_lab/common/collect/PersistentDeque.java
```diff
@@ -0,0 +1,190 @@
...
+ * the deque).
+ *
+ * <p>Currently, it is only possible to create an empty deque and then add new elements one
+ * at a
+ * time.
+ *
+ * @param <T> type of elements to be stored in deque
+ */
+@Immutable(containerOf = "T")
+public final class PersistentDeque<T> implements PersistentDequeInterface<T> {
```
baierd: It makes sense to name your implementation in a ways that explains the core concept. Example: `PersistentBalancingDoubleListDeque`

src/org/sosy_lab/common/collect/PersistentDequeInterface.java
baierd: The interface itself should reflect the actual name of the data-structure (i.e. `PersistentDeque`), as it is already an interface (you would not include the string `method` in the name of a method ;D ). Developers should decide whether they want "some" `Deque` or need the specialized properties of a `PersistentDeque`. Similarly, developers can choose to include more type info by using the implementation if that's necessary. The implementations should then choose names that signals details about their implementation. Example: `List` is the interface used for the `LinkedList` implementation.

src/org/sosy_lab/common/collect/PersistentDeque.java
```diff
@@ -0,0 +1,190 @@
...
+   * Removes element at the bottom of the deque from deque.
+   *
+   * @return deque instance after bottom element has been removed
+   */
+  @Override
+  public PersistentDeque<T> deleteBottom() {
+    return new PersistentDeque<>(top, bottom.tail()).rebalanceDeque();
+  }
+
+  private PersistentDeque<T> rebalanceDeque() {
```
baierd: It makes sense to document the invariant that has to hold for this method and for the `split()` method. Also, you can use assertions to actually check them.

src/org/sosy_lab/common/collect/PersistentDeque.java
```diff
@@ -0,0 +1,190 @@
...
+ * other list is split up into top and bottom to further guarantee access to both ends of
+ * the deque).
+ *
+ * <p>Currently, it is only possible to create an empty deque and then add new elements one
+ * at a
+ * time.
+ *
+ * @param <T> type of elements to be stored in deque
+ */
+@Immutable(containerOf = "T")
```
baierd, line 26: The class is declared to be immutable, is actually immutable, but it is still using/allowing mutable methods.

baierd (changes requested):
Good idea with the 2 `PersistentLinkedList`s!

Some formalities need to be improved (namings, inheritance of the interface from Javas definition of Deque etc.). But most of it is just copy/pasting and renamings.

I did not yet check the balancing (invariant).

src/org/sosy_lab/common/collect/PersistentDequeInterface.java
baierd: You actually had a reasoning why you did not import `java.util.Deque` in your description. You should still simply inherit `java.util.Deque`, and if you don't support methods etc., you just throw an `UnsupportedOperationException` with a small explanation.

Note: you actually should support at least the core methods/concepts in `java.util.Deque`, like the iterators.

.gitignore
c-m-elliott: I had totally forgotten I had put that there. I never meant for it to be merged to main, I just wanted a temporary fix for the problem. Please will you let me know when you've updated the settings so I should be okay to remove it?
Thanks for taking care of that!

src/org/sosy_lab/common/collect/PersistentDequeInterface.java
c-m-elliott: I'm not sure if I included in my reasoning that some of the methods required return types that I didn't think matched up with what would make sense for a persistent version. I.e. pop returns the element that was removed, which would leave no way to return the new, updated instance of the deque. Is that one of the cases where you would like it to throw an exception and create a different method that actually works instead?

.gitignore
baierd: This change will not help you anyway, so please just remove it. Gitignore only ignores new files, not file changes.

src/org/sosy_lab/common/collect/PersistentDeque.java
c-m-elliott: I was a bit iffy about that too, but other options seemed bulky. As long as the rebalancing works correctly and is called where necessary, bottom should only contain at most one element if top.head() threw an exception. I will admit though that this logic relies on the rest of the code being sound, or it could cause mistakes...
I will try figure out a nicer way of doing it.

src/org/sosy_lab/common/collect/PersistentDequeInterface.java
c-m-elliott: I agree, I just didn't know what to call the implementing class other that 'PersistentDeque' at the beginning. I had planned to change it in the end but clearly, I forgot. Thanks for the reminder.

src/org/sosy_lab/common/collect/PersistentDeque.java
c-m-elliott: I like that the name you suggested is very descriptive of how the deque is achieved. Is it a bit long though, or will it be fine as it is? Otherwise, I could try and come up with something slightly shorter.

.gitignore
c-m-elliott: Will do!

src/org/sosy_lab/common/collect/PersistentDequeInterface.java
baierd: That is why we throw for all modifying methods in immutable data-structures. In your version you have the worst of both worlds, they look like they are mutable, but are not.
You can take a look at `AbstractImmutableMap` for details for example.

Usually we then introduce new methods following the copy-then-modify approach, for example `addAndCopy()` instead of `add()`. `addAndCopy()` then returns a new version of the current data-structure that is also immutable and persistent, with the new element added.
You can take a look at `PersistentMap` and its implementation for details for example.

src/org/sosy_lab/common/collect/PersistentDeque.java
baierd: OK. I'll take a closer look at the re-balancing once the formalities/API renamings etc. are done.

src/org/sosy_lab/common/collect/PersistentDeque.java
baierd: If your name does not exceed 80 characters, everything is OK.

.gitignore
c-m-elliott: Has been taken care of.

src/org/sosy_lab/common/collect/PersistentDeque.java
c-m-elliott: Have renamed it to what you suggested.

src/org/sosy_lab/common/collect/PersistentDequeInterface.java
c-m-elliott: The interface has been renamed as well.

src/org/sosy_lab/common/collect/PersistentBalancingDoubleListDequeTest.java
```diff
@@ -0,0 +1,102 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package org.sosy_lab.common.collect;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class PersistentBalancingDoubleListDequeTest {
```
baierd: As long as this test class only uses methods defined by the `PersistentDeque` interface, and works purely in a black-box fashion, it can be named `PersistentDequeTest`.

baierd (changes requested):

src/org/sosy_lab/common/collect/PersistentDeque.java
c-m-elliott: Done.

src/org/sosy_lab/common/collect/PersistentDeque.java
c-m-elliott: This should be resolved as all methods returning null do so because they were specified that way in java.util.Deque.

src/org/sosy_lab/common/collect/PersistentDeque.java
baierd: You don't need to list methods that are just handed through, like `Iterator<T> iterator();`.

Overall this looks much better! Thank you!

src/org/sosy_lab/common/collect/PersistentBalancingDoubleListDequeTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+  PersistentBalancingDoubleListDeque<Object> emptyDeque;
```
baierd: I am questioning whether plain empty `Object`s, and constant "full" (can this even be "full"?) and size 1 Deques are a good idea to test this data-structure in general.

This can be simplified by removing all of the class variables, as they seem like they can be set up quickly and targeted in actual tests.

Also, using proper, established types instead of plain `Object`s should be highly considered!

baierd (changes requested):

src/org/sosy_lab/common/collect/PersistentDeque.java
baierd: Control-flow with exceptions should be avoided as much as possible (it is slow, hard to read/debug etc.). Why don't you simply check whether top/bottom are empty before calling `head()`?

There are multiple occurrences of this behavior in this class.

c-m-elliott: You're right, they could have been avoided altogether. I've removed them all and altered the methods accordingly.

PR #55, opened 2026-04-19
Throw NoSuchElementException for empty PersistentLinkedList querying an element

### LINKED ISSUE #54: PersistentLinkedList throws IllegalStateException instead of NoSuchElementException for operations on an empty list
Our `PersistentLinkedList` throws `IllegalStateException`s instead of `NoSuchElementException`s for operations on an empty lists.

baierd:
Throw NoSuchElementException for empty PersistentLinkedList querying an element instead of IllegalStateException.

Closes #54.


PhilippWendler:
Good question whether we want this. There is no interface that these two methods implement and which would specify what method we need to throw.

On the one hand, there is a consistency argument with methods from `List`.

On the other hand, especially for `tail()` it feels more like a `IllegalStateException`. After all, it does not throw because there are no elements: Calling `tail()` on a one-element list returns an empty list. `tail()` throws because the request for the tail of an empty list is not meaningful. For `head()` a `NoSuchElementException` seems suitable, but we probably also do not want a different exception for `head()` and `tail()`.

So yes, we could merge, but we also could just keep it as is for simplicity?

baierd:
I only noticed by accident and would slightly tend towards changing it.
The proposed change is how Javas `LinkedList` behaves, so it would increase consistent behavior by a tiny amount.


Note: the naming of these methods is also not consistent with how Java names them. E.g. `head()` -> `getFirst()`. But that's a breaking change, hence why i didn't propose a change yet without asking.

PhilippWendler:
For `head()`, ok.

But for `tail()`, `NoSuchElementException` does not match.

Do you prefer the inconsistency between `head()` and `tail()`?

baierd:
I don't care about merging this or changing it again or no changing it at all, I want a fitting solution.

I am also unsure whether `NoSuchElementException` does fit `tail()` well. Its [documentation](https://docs.oracle.com/javase/8/docs/api/java/util/NoSuchElementException.html) says "Thrown by various accessor methods to indicate that the element being requested does not exist.", which seems to fit from its description. I see that "element" might be confusing here though due to it returning a list and not a list element.

I am also unsure whether `IllegalStateException` fits in a better way, as the "state" is not illegal in general, just for this method call. I understand that [the documentation](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalStateException.html) says "Signals that a method has been invoked at an illegal or inappropriate time. ..." which covers our use case, but i feel like it might be misunderstood in this case.

(I wish [HeadlessException](https://docs.oracle.com/javase/8/docs/api/java/awt/HeadlessException.html) had a tail counterpart.... :rofl: )

Java does not provide a blueprint for this in other methods, so i looked up what Scala does, and they throw `UnsupportedOperationException` for an empty tail when calling `tail()`, and a `NoSuchElementException` for `head()` in an empty list. 

But this seems to discourage use, as nobody wants to catch an exception in such cases. Another proposed solution is returning an empty list instead, with the main issue seemingly being consistency with how other methods behave ([Source](https://github.com/scala/scala3/issues/22258)). But Scala also provides API like `headOption()` that returns an `Optional`, and i really like this, but this is clearly optional additional API.

Kotlin avoids the problem (as Java does) by simply not providing a `tail()` method.

So actually i would go with what Scala does here and throw two distinct exceptions, and only switch `head()` to `NoSuchElementException`, as it fits naturally.

PhilippWendler:
If you care enough to change `head()` while keeping `tail()`, ok, we can do this.

baierd:
I changed it, and i updated the JavaDoc.

I also added 2 "safe" methods that return an optional.

PhilippWendler:
Are the new methods useful, do you have a case where you actually use them?

I am not too fond of adding methods where the usefulness is not clear, because it clutters the API and locks us in in the future. (And there are not even tests.)

baierd:
Fair point. In Java you need explicit checking either way, so it is not really useful. I removed them.

PhilippWendler (approved):

PR #61, opened 2026-05-13
60 add avl tree as an alternative backend data structures for copy efficent persistent data structures

### LINKED ISSUE #60: Add AVL Tree as an Alternative Backend Data-Structures for Copy-Efficent (Persistent) Data-Structures
An [Adelson-Velsky and Landis (AVL) tree](https://en.wikipedia.org/wiki/AVL_tree) is a self-balancing Binary Search Tree with distinct balancing to our current default (LL)RB tree. AVL trees have equal amortized and worst case time complexity to RB trees, with differences in several use-cases due to their distinct structure and balancing. AVL trees are height-balanced, similar to RB trees, but AVL trees are often faster than RB trees for lookup-intensive applications due to AVL trees having stricter balancing than RB trees. This balancing might be more expensive compared to RB trees though.

kaizhx:
## Summary

Adds an initial path-copying persistent AVL tree map implementation based on the existing `PathCopyingPersistentTreeMap` which is based on a LLRB tree. 

## Changes 

Changes to the existing tree map implementation focus on replacing LLRB specific logic with their AVL counterparts:

- Added `PathCopyingPersistentAvlTreeMap`
- Replaced color metadata stored in nodes with AVL height metadata.
- Added AVL balance factor calculation based on subtree heights.
- Replaced LLRB invariant restoration with AVL rebalancing via `rebalance()`.
- Implemented trinode restructuring for simplified rebalancing logic via `restructure()`.
(See https://www.csl.mtu.edu/cs2321/www/newLectures/20_AVL_Tree.html for more information on restructuring algorithm)
- Updated `checkAssertions()` to validate AVL height and balance invariants instead of LLRB invariants.
- Reworked insertion via `putAndCopy()` to use AVL rebalancing.
- Reworked deletion via `removeAndCopy()` to use standard BST deletion followed by AVL rebalancing.
- Implemented Browns improved deletion algorithm for AVL trees in `restructure()` for the two child deletion case, choosing the predecessor/successor replacement based on the larger subtree height.
(See https://arxiv.org/pdf/2406.05162 for Browns improved deletion algorithm)

## Reused structures

The implementation intentionally keeps most of the existing persistent tree map structure unchanged where logic is not tied to the LLRB tree implementation and only requires the existence of a binary search tree.

- Immutable node structure
- Empty map handling
- Persistent path-copying style
- Entry iteration
- helper functions (min/max node, lookup, bounds)
- Bounded views support (e.g. subMap, headMap, tailMap)

### DIFF
--- a/src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMap.java
+++ b/src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMap.java
@@ -1,0 +1,1478 @@
+// This file is part of SoSy-Lab Common,
+// a library of useful utilities:
+// https://github.com/sosy-lab/java-common-lib
+//
+// SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package org.sosy_lab.common.collect;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.base.Verify.verify;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Iterators;
+import com.google.common.collect.UnmodifiableIterator;
+import com.google.errorprone.annotations.Immutable;
+import com.google.errorprone.annotations.Var;
+import com.google.errorprone.annotations.concurrent.LazyInit;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.io.Serializable;
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.ArrayDeque;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.Deque;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.NavigableMap;
+import java.util.NavigableSet;
+import java.util.Objects;
+import java.util.SortedMap;
+import java.util.function.BinaryOperator;
+import java.util.function.Function;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * This is an implementation of {@link PersistentSortedMap} that is based on the Adelson-Velsky and
+ * Landis (AVL) tree and path copying for copy-efficient persistency.
+ *
+ * @param <K> The type of keys.
+ * @param <V> The type of values.
+ */
+@Immutable(containerOf = {"K", "V"})
+@SuppressFBWarnings(value = "SING_SINGLETON_IMPLEMENTS_SERIALIZABLE", justification = "false alarm")
+public final class PathCopyingPersistentAvlTreeMap<
+    K extends Comparable<? super K>, V extends @Nullable Object>
+    extends AbstractImmutableSortedMap<K, V> implements PersistentSortedMap<K, V>, Serializable {
+
+  private static final long serialVersionUID = 1041711151457528188L;
+
+  @SuppressWarnings("unused")
+  @SuppressFBWarnings(
+      value = "EQ_DOESNT_OVERRIDE_EQUALS",
+      justification = "Inherits equals() according to specification.")
+  @Immutable(containerOf = {"K", "V"})
+  private static final class Node<K, V extends @Nullable Object>
+      extends SimpleImmutableEntry<K, V> {
+
+    private static final long serialVersionUID = -7393505826652634501L;
+
+    private final @Nullable Node<K, V> left;
+    private final @Nullable Node<K, V> right;
+    private final byte height;
+
+    // Leaf node
+    Node(K pKey, V pValue) {
+      this(pKey, pValue, null, null);
+    }
+
+    // Any node
+    Node(K pKey, V pValue, @Nullable Node<K, V> pLeft, @Nullable  Node<K, V> pRight) {
+      super(pKey, pValue);
+      left = pLeft;
+      right = pRight;
+      height = (byte) (1 + Math.max(getHeight(left), getHeight(right)));
+    }
+
+    boolean isLeaf() {
+      return left == null && right == null;
+    }
+
+    static byte getHeight(@Nullable Node<?, ?> n) {
+      return n == null ? -1 : n.height;
+    }
+
+    static byte getBalanceFactor(Node<?, ?> n) {
+      return (byte) (getHeight(n.left) - getHeight(n.right));
+    }
+
+    // Methods used for path copying
+
+    @SuppressWarnings("ReferenceEquality") // cannot use equals() for check whether tree is the same
+    Node<K, V> withLeftChild(Node<K, V> newLeft) {
+      if (newLeft == left) {
+        return this;
+      } else {
+        return new Node<>(getKey(), getValue(), newLeft, right);
+      }
+    }
+
+    @SuppressWarnings("ReferenceEquality") // cannot use equals() for check whether tree is the same
+    Node<K, V> withRightChild(Node<K, V> newRight) {
+      if (newRight == right) {
+        return this;
+      } else {
+        return new Node<>(getKey(), getValue(), left, newRight);
+      }
+    }
+
+    static int countNodes(@Nullable Node<?, ?> n) {
+      if (n == null) {
+        return 0;
+      }
+      return countNodes(n.left) + 1 + countNodes(n.right);
+    }
+  }
+
+  // static creation methods
+
+  private static final PathCopyingPersistentAvlTreeMap<?, ?> EMPTY_MAP =
+      new PathCopyingPersistentAvlTreeMap<String, Object>(null);
+
+  @SuppressWarnings("unchecked")
+  public static <K extends Comparable<? super K>, V extends @Nullable Object>
+  PersistentSortedMap<K, V> of() {
+    return (PersistentSortedMap<K, V>) EMPTY_MAP;
+  }
+
+  public static <K extends Comparable<? super K>, V extends @Nullable Object>
+  PersistentSortedMap<K, V> copyOf(Map<K, V> map) {
+    checkNotNull(map);
+
+    if (map instanceof PathCopyingPersistentAvlTreeMap<?, ?>) {
+      return (PathCopyingPersistentAvlTreeMap<K, V>) map;
+    }
+
+    @Var PersistentSortedMap<K, V> result = of();
+    for (Map.Entry<K, V> entry : map.entrySet()) {
+      result = result.putAndCopy(entry.getKey(), entry.getValue());
+    }
+    return result;
+  }
+
+  /**
+   * Return a {@link Collector} that accumulates elements into a {@link
+   * PathCopyingPersistentAvlTreeMap}. Keys and values are the result of the respective functions. If
+   * duplicate keys appear, the collector throws an {@link IllegalArgumentException}.
+   */
+  public static <T, K extends Comparable<? super K>, V extends @Nullable Object>
+  Collector<T, ?, PersistentSortedMap<K, V>> toPathCopyingPersistentAvlTreeMap(
+      Function<? super T, ? extends K> keyFunction,
+      Function<? super T, ? extends V> valueFunction) {
+    return toPathCopyingPersistentAvlTreeMap(
+        keyFunction,
+        valueFunction,
+        (k, v) -> {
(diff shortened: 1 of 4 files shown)


src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+import java.util.TreeMap;
+import junit.framework.JUnit4TestAdapter;
+import junit.framework.TestSuite;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+@SuppressWarnings("MemberName")
+public class PathCopyingPersistentAvlTreeMapTest {
```
baierd, line 38: The optimal way of setting up the tests would be to have all tests for `PersistentMap`s (i.e. tests for maps in general) in a single class and execute it as a parameterized test for all of our map implementations. JUnit 4 provides this feature relatively easily. An example implementation of ours can be found [here](https://github.com/sosy-lab/java-smt/blob/master/src/org/sosy_lab/java_smt/test/SolverBasedTest0.java).

I would suggest to add a new class that handles parameterized testing of `PersistentMap`s (similar to [SolverBasedTest0](https://github.com/sosy-lab/java-smt/blob/master/src/org/sosy_lab/java_smt/test/SolverBasedTest0.java)) and then  extend `PathCopyingPersistentTreeMapTest` with it + tests from this class that are not yet present in `PathCopyingPersistentTreeMapTest`. (The name `PathCopyingPersistentTreeMapTest` indicates that it runs all kinds of tree based persistent maps, instead of just RB-Trees anyways)
Test cases that use internal API of your AVL-Tree should remain here (if any).

Tests that test the (currently 2, so parameterizing does not make sense yet) maps against each other should be moved to a dedicated test-class with a proper name.

But there are some other classes that only test on `PathCopyingPersistentTreeMap`, e.g. `CopyOnWriteSortedMapTest`, that should also be parameterized.

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+  private @Nullable NavigableMap<String, String> reference;
+
+  @Before
+  public void setUp() {
+    map = PathCopyingPersistentAvlTreeMap.of();
+  }
+
+  @After
+  public void tearDown() {
+    map = null;
```
baierd, line 80: This is not needed. Since your setup method uses `@Before`, it is guaranteed that you start with a fresh map for each test. This would also allow you to remove the `@Nullable` annotation (that you don't want to use anyway. Avoiding `null` and treating it as an error should be the default nowadays)

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+                CollectionFeature.SERIALIZABLE_INCLUDING_VIEWS,
+                CollectionSize.ANY)
+            .createTestSuite());
+
+    return suite;
+  }
+
+  private @Nullable PersistentSortedMap<String, String> map;
+  private @Nullable PersistentSortedMap<String, String> llrb;
+  private @Nullable NavigableMap<String, String> reference;
```
baierd, line 71: Since you don't use these 2 in your setup, you should delete them and use local variables instead.

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+  private void put(@Nullable String context, String key, String value) {
+    PersistentSortedMap<String, String> oldMap = map;
+    int oldMapSize = oldMap.size();
+    String oldMapStr = oldMap.toString();
+
+    map = map.putAndCopy(key, value);
+
+    try {
+      ((PathCopyingPersistentAvlTreeMap<?, ?>) map).checkAssertions();
+    } catch (IllegalStateException e) {
```
baierd, line 96: Why not just throw/not catch the `IllegalStateException`?

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+  @After
+  public void tearDown() {
+    map = null;
+  }
+
+  private void put(String key, String value) {
+    put(null, key, value);
+  }
+
+  private void put(@Nullable String context, String key, String value) {
```
baierd, line 87: You could improve the naming of this method and add what it does. Also, you should add some JavaDoc to explain the purpose of this method.

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+        assertThat(map).isNotEqualTo(oldMap);
+      }
+
+    } else {
+      assertThat(map).hasSize(oldMap.size() + 1);
+      assertThat(map).isNotEqualTo(oldMap);
+    }
+  }
+
+  private void remove(@Nullable String context, String key) {
```
baierd, line 122: You could improve the naming of this method and add what it does. Also, you should add some JavaDoc to explain the purpose of this method.

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+  }
+
+  private void remove(String key) {
+    remove(null, key);
+  }
+
+  @Test
+  public void testEmpty() {
+    assertThat(map.toString()).isEqualTo("{}");
+    assertThat(map).isEmpty();
```
baierd, line 158: I would argue that it makes sense to start this test with this method ;D

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+    assertThat(map).hasSize(0);
+    assertThat(map.hashCode()).isEqualTo(0);
+  }
```
baierd, line 161: You should also test the behavior of methods like `remove*` on empty maps.

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+
+  private void putABCD() {
```
baierd, line 163: This method should have some documentation and improved naming. Also, i recommend checking that the map is actually empty before adding elements (as you expect this to be the case in your checks ;D).

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+    assertThat(map.firstKey()).isEqualTo("a");
+    assertThat(map.lastKey()).isEqualTo("c");
+
+    put("d", "4");
+    assertThat(map.toString()).isEqualTo("{a=1, b=2, c=3, d=4}");
+    assertThat(map.firstKey()).isEqualTo("a");
+    assertThat(map.lastKey()).isEqualTo("d");
+  }
+
+  private void removeDCBA() {
```
baierd, line 185: I recommend not exporting all of your testing into methods, as it makes reading/modifying the tests way harder! It makes sense to export some common setups like `putABCD()` or operations like `put()` above, but i would advice not to export this or other modifications that don't need large amounts of code after the setup.

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+
+  private void removeABCD() {
+    remove("a");
+    remove("b");
+    remove("c");
+    remove("d");
+  }
+
+  @Test
+  public void testRight() {
```
baierd, line 222: Please add some documentation to your tests and/or improve the naming of them (method names up to 80-100 characters are totally fine!). Currently it is nearly impossible to see what they do immediately, due to your method names being not expressive enough and/or delegation of the actual test to other methods (that are also not named very expressively ;D).

This applies to this class in general.

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+    ((PathCopyingPersistentAvlTreeMap<?, ?>) map).checkAssertions();
+    ((PathCopyingPersistentTreeMap<?, ?>) llrb).checkAssertions();
+  }
+
+  private void checkPartialMaps(NavigableMap<String, String> comparison, Random rnd) {
+    String key1 = Integer.toString(rnd.nextInt());
+    String key2 = Integer.toString(rnd.nextInt());
+
+    checkEqualTo(
+        comparison.tailMap(key1, /* inclusive= */ true), map.tailMap(key1, /* pInclusive= */ true));
```
baierd, line 436: Whats the purpose of the comments in this method?

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+    map = map.putAndCopy("a", "a").putAndCopy("b", "b").putAndCopy("c", "c");
+
+    NavigableMap<String, String> submap = map.subMap("aa", "c");
+    assertThat(submap).containsExactly("b", "b");
+
+    // The bounds of further submap calls may be at most those of the original call
+    @Var NavigableMap<String, String> subsubmap = submap.subMap("aaa", true, "c", false);
+    assertThat(subsubmap).containsExactly("b", "b");
+
+    subsubmap = submap.subMap("aa", /* fromInclusive= */ true, "bb", /* toInclusive= */ false);
```
baierd, line 515: Whats the purpose of these comments: `fromInclusive` etc?

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
```diff
@@ -0,0 +1,538 @@
...
+                MapFeature.ALLOWS_NULL_VALUES,
+                CollectionFeature.KNOWN_ORDER,
+                CollectionFeature.SERIALIZABLE_INCLUDING_VIEWS,
+                CollectionSize.ANY)
+            .createTestSuite());
+
+    return suite;
+  }
+
+  private @Nullable PersistentSortedMap<String, String> map;
```
baierd, line 69: Why only test `<String, String>`?

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMap.java
```diff
@@ -0,0 +1,1484 @@
+// This file is part of SoSy-Lab Common,
+// a library of useful utilities:
+// https://github.com/sosy-lab/java-common-lib
+//
+// SPDX-FileCopyrightText: 2007-2020 Dirk Beyer <https://www.sosy-lab.org>
```
baierd, line 5: `2007-2026`

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMap.java
```diff
@@ -0,0 +1,1484 @@
...
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * This is an implementation of {@link PersistentSortedMap} that is based on the Adelson-Velsky and
+ * Landis (AVL) tree and bottom-up path copying for copy-efficient persistency.
+ *
+ * <p>
+ * TODO: Finish documentation
```
baierd, line 46: ;D

src/org/sosy_lab/common/collect/PathCopyingPersistentAvlTreeMapTest.java
baierd, line 38: Did you test all operations (e.g. rotation cases) of AVL trees?

If not, i would suggest to map out cases for which you can trigger all of them and add them here.

baierd (changes requested):

PR #62, opened 2026-06-25
49 add union find data structure

### LINKED ISSUE #49: Add Union-Find Data-Structure
[Union-Find (also called Disjoint-set)](https://en.wikipedia.org/wiki/Disjoint-set_data_structure) is a data-structure not found in Java per default, and neither our dependencies (e.g. Guava). It stores a collection of disjoint sets, and is useful when it comes to graph-based operations/implementations. It defines 2 key operations `find(x)`, returning a canonized element of the set that x is in, and `union(x, y)`, merging the sets of x and y. Deleting elements is often useless in practice, especially if there are immutable versions available, so this operation is not a focus for optimization, but should be included as far as possible. Since there are graph-based applications/implementations in CPAchecker (i.e. SMGs), this data-structure can be helpful for us. Hence we should add it.

Union-Find can be optimized in many ways.  Insertion can have constant (`O(1)`) worst case time complexity. Search up to inverse-Ackermann amortized worst case time-complexity (`O(α(n))`). Meaning not every search operation is within `O(α(n))` due to re-balancing (see below), but re-balancing causes subsequent operations to be faster, leading to an amortized worst case time-complexity of `O(α(n))`. In practice this can lead to `find()` and `union()` only needing near constant time.

Optimally we want a mutable, immutable and a persistent version of union-find.
Mutable and immutable implementations may also be persistent for union-find, as the cost is minimal in practice.

Persistence can be achieved with path compression using trees by utilizing either union by rank or weight, leading to search operations having `O(log n)` worst case time complexity. This is often extended by splitting by rank or weight, reducing needed (amortized) time further.

Alternatively also by tree compression (up to an optimal worst case time of O(log n / log log n)) during union [[3](https://dl.acm.org/doi/10.1145/301250.301383)].

Often a tree based disjoint-set forest is used in practice. However, these are commonly parent pointer trees. We need to explore whether our LLRB-tree can still be used, or is a good idea (due to its auto-balancing).

In any case; the implementation should first define a new interface for the common data-structure, implementing commonalities from Javas interfaces that make sense, followed by sub-interfaces like an ordered version.

Then, an "easy" (non-optimized) implementation that is easily verifiable as correct should be added (there are lots of examples online. Be careful with licenses! Either we need to import the original license of used code, or better use a license free one).
This non-optimized union-find should then be used to show base principles of the implementation, define tests upon etc. This is particularly helpful to evaluate against (and potentially as test-oracle).

More complex implementations can then either be based upon the non-optimized union-find or more complex ideas.

The iteration order in sets, as well as over sets should be consistent!

c-m-elliott:
So far, this includes the interfaces UnionFind, SortedUnionFind, PersistentUnionFind and PersistentSortedUnionFind as well as the abstract classes AbstractImmutableUnionFind and AbstractImmutableSortedUnionFind. It also contains a naive implementation of SortedUnionFind based on a HashMap of TreeSets (SortedTreeSetUnionFind). There is a set of tests (SortedUnionFindTest), but there seems to be some kind of problem regarding null values which is causing the JDK21 PackageSanityTest to fail. This in particular is something I'd appreciate input on. Thanks!


baierd:
TODO (WIP): make a concrete list of all things that we expect this PR to have with some examples (e.g. some interesting test-cases). Smaller work-packages are preferred.

# List of wants and goals for this MR

Union-Find commonly only defines two methods: 
- the `find()` method. It has a single argument and returns the representative member of the set of elements that the argument belongs to.
- the `union()` method. It takes two arguments, merging the two sets that each argument belongs to. In mutable data-structures the return type is `void`. Persistent and immutable versions deprecate this call. Persistent implementations add a new method `unionAndCopy()` returning a new copy of the initial Union-Find, with the two sets merged. The initial data-structure of the persistent implementation is unchanged after returning the new modified copy.

Another method `getAllSets()` or `getAllSubSets()` returning a set of all sets inside the Union-Find might be helpful and should also be added.

We should discuss whether we want size method(s), e.g. for the amount of sets or elements in all sets.

Sometimes a `addition` method is used, but i don't see any benefit for this, so we will not add it.

## Interfaces (or abstract classes)
- [ ] Interface for mutable Union-Find
- [ ] Interface for ordered mutable Union-Find based on the interface for mutable Union-Find.
- [ ] Interface for immutable Union-Find based on the interface of mutable Union-Find.
- [ ] Interface for ordered immutable Union-Find based on the interfaces of immutable Union-Find and ordered Union-Find. Should multi-inheritance be a problem, you can switch out one of the 2 source interfaces for an abstract class. Ordered mutable Union-Find seems to be the better choice for this at a first glance, but you can evaluate this for yourself.
- [ ] Interface for persistent Union-Find based on the interface for immutable Union-Find.
- [ ] Interface for ordered persistent Union-Find based on the interface for persistent Union-Find and ordered immutable Union-Find. (See ordered immutable Union-Find above for the multi-inheritance problem you might face.)

## Implementations
- [ ] Mutable Union-Find
- [ ] Immutable Union-Find
- [ ] Persistent Union-Find

Union-Find data-structures are usually based on trees and can be implemented using commonly existing implementations (e.g. our `PathCopyingPersistentTreeMap` that is based on a RB-Tree). The problem is, these can not be optimized with the techniques below. A optimized tree for Union-Find needs to be a [Parent-Pointer-Tree](https://en.wikipedia.org/wiki/Parent_pointer_tree), i.e. a tree with its nodes storing only a single pointer towards their parents.

It makes sense to start with a "simple" implementation first, e.g. based on our `PathCopyingPersistentTreeMap` and/or a naive parent-pointer-tree without optimizations. This would also allow us to evaluation more optimized versions against naive implementations later on!

- [ ] Ordered mutable Union-Find
- [ ] Ordered immutable Union-Find
- [ ] Ordered persistent Union-Find

Ordered parent-pointer-tree implementations seem to require a proper merge of the trees in most cases. It makes sense to focus on the unordered variants first. TODO: investigate further.

Note: a "unordered" data-structure implementation can sometimes be derived from an ordered. An example of this is our `PathCopyingPersistentTreeMap`, which is always ordered.

## Tests
- [ ] Mutable Union-Find tests. This should test the behavior for Union-Find. For example:
  - [ ] TODO: add examples
- [ ] Ordered mutable Union-Find tests. Extend the mutable Union-Find tests by tests for the new API and ordering.
- [ ] Immutable Union-Find tests. Simply translate the tests of mutable Union-Find to this.
- [ ] Ordered immutable Union-Find tests. Similar to ordered mutable Union-Find tests.
- [ ] Persistent Union-Find tests. Simply translate the tests of (im)mutable Union-Find to this.
- [ ] Ordered persistent Union-Find tests. Similar to ordered mutable Union-Find tests.

I advice to "extend" tests only by the new features introduced in each implementation, and re-use existing tests as far as possible. E.g. ordered Union-Find test classes can just extend their unordered test-classes.

## Documentation

All public methods and classes should be documented with their purpose, behavior, constraints, and edge cases, i.e. what callers may rely on.
Any code that would prompt questions from a developer who is proficient in Java but unfamiliar with the project should be considered non-trivial and documented.

## (Runtime) Optimization
- [ ] Path Compression (Path compression links set members to their root, making `find` operations more efficient.) 
- [ ] union by rank. This requires a background tree that stores its rank (i.e. the tree heights upper bound) in each root.
- [ ] union by size. This requires a size-weighted tree as the background data-structure.

Our goal is to use at least one `union` optimization together with path compression. This should lead to a Union-Find implementation that is close to the best worst case time complexities possible if both are used. For `find`, the best possible worst-case time-complexity in this case is `O(log n / log log n​)`, while `union` remains O(α(n)) (amortized).

Space complexity is usually O(n) and does not need to be optimized.
The Guava trick of actually having two data-structures saving all elements to reduce run-time for certain operations can be used of course. 

## Evaluation

- We want to use [BenchExec](https://github.com/sosy-lab/benchexec) to run the experiments, as it ensures reliable benchmarking and resource measurement.
- We focus on run-time (CPU-time) when evaluating this. Memory usage should be roughly confirmed to be O(n).
- We want to evaluate all additions using common workload (adding elements, removing elements, checking whether elements are contained and in which sub-set).
- Furthermore, we want to evaluate all optimized versions against some primitive union-find implementation without optimizations, as well as a matching Set of Sets.

c-m-elliott:
@baierd could you please take a look at ParentPointerTreeUnionFind? It uses ParentPointerTree and TreeNode which both don't implement any standard interfaces at the moment. For example, TreeNode could implement Object and ParentPointerTree could have Collection. They would mean quite a few additional methods which I don't need for the union-find. I am wondering whether they might be valuable further down the line (in case of potential future uses) or just add bulk without real benefit. What do you think?

baierd:
> For example, TreeNode could implement Object 

That would make no sense. Every concrete instance of a class is automatically a object.

> and ParentPointerTree could have Collection.

Do you really want to implement everything from `Collection` just for your Union-Find?

You don't need to implement a interface if they are not helpful.

c-m-elliott:
Thanks, then I will leave them as they are. I don't see myself needing Collection features with what I have planned at the moment.

src/org/sosy_lab/common/collect/union_find/tests/ImmutableParentPointerTreeUnionFindSortednessTest.java
```diff
@@ -0,0 +1,193 @@
...
+import static com.google.common.truth.Truth.assertThat;
+
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.NavigableSet;
+import org.junit.Test;
+import org.sosy_lab.common.collect.union_find.ImmutableSortedParentPointerTreeUnionFind;
+import org.sosy_lab.common.collect.union_find.ParentPointerTreeUnionFind.UnionType;
+
+public class ImmutableParentPointerTreeUnionFindSortednessTest {
```
baierd, line 20: The name can be simplified:
```suggestion
public class ImmutableSortedUnionFindTest {
```

Possibly even more, depending on whether you test its immutability in this class or not (which would be totally legit).
Specify type information should only be included if you actually test its specifics.
As far as i can see, this only tests sortedness, hence neither immutability nor the implementation specifics of the `UnionFind` (i.e. that it is a `ParentPointerTree` or `Tree`) matter here.

src/org/sosy_lab/common/collect/union_find/tests/ImmutableParentPointerTreeUnionFindSortednessTest.java
```diff
@@ -0,0 +1,193 @@
...
+
+    Collection<? extends NavigableSet<T>> subsets = pSortedUnionFind.getAllSubsets();
+
+    assertThat(subsets).hasSize(1);
+
+    return subsets.iterator().next();
+  }
+
+  @Test
+  public void testGetAllSubsets_elementsAddedInAscendingOrder_remainSorted() {
```
baierd, line 46: Our Java naming convention ([see Google style guide](https://google.github.io/styleguide/javaguide.html)) uses camel case, not snake case.
Also, tests should add the suffix `*Test` always at the end of their name.

This should be considered in the entire MR.

src/org/sosy_lab/common/collect/union_find/ImmutableParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * @param <T> type of elements added to the Union-Find.
+ */
+@Immutable(containerOf = "T")
+public class ImmutableParentPointerTreeUnionFind<T> extends AbstractImmutableUnionFind<T> {
+
+  // tree nodes are not immutable but only used internally and never mutated after creation
+  // immutable tree nodes would make conversion during build() difficult and time-consuming
+  @SuppressWarnings("Immutable")
+  private final ImmutableMap<T, AbstractTreeNode<T>> allNodes;
```
baierd, line 38: This is not a immutable `UnionFind` implementation. While your Map is immutable, the nodes, and therefore the tree, are not immutable. Modifying a previous copy also modifies this copy and vice versa.

src/org/sosy_lab/common/collect/union_find/ImmutableParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * Provides a {@link Collection} containing all current subsets.
+   *
+   * @return {@link Collection} containing all current subsets
+   */
+  @Override
+  public Collection<? extends Set<T>> getAllSubsets() {
+
+    Map<T, Set<T>> allSubsets = new HashMap<>();
```
baierd, line 85: This is a immutable class. Why not return a immutable collection here?

src/org/sosy_lab/common/collect/union_find/ImmutableParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,140 @@
...
+   *
+   * @param <T> type of elements added to the Union-Find
+   */
+  public static final class Builder<T> extends AbstractImmutableParentPointerTreeBuilder<T> {
+
+    private Builder(UnionType pUnionType) {
+      super(pUnionType);
+    }
+
+    public static <T> AbstractImmutableParentPointerTreeBuilder<T> getBuilder(
```
baierd, line 130: The static constructor method for `Builder` should be part of the public interface of the immutable classes, and this method should be moved there.


The idea of a `Builder` is that you automatically get the correct class when requesting it through the static constructor method. But here you hard-coded them, making this impossible. A user needs to know the details of the class at all points when programming, making this inconvenient.

src/org/sosy_lab/common/collect/union_find/AbstractImmutableUnionFind.java
```diff
@@ -0,0 +1,28 @@
...
+import com.google.errorprone.annotations.DoNotCall;
+import com.google.errorprone.annotations.Immutable;
+
+/**
+ * An abstract class for immutable Union-Find implementations.
+ *
+ * @param <T> type of elements added to the Union-Find
+ */
+@Immutable(containerOf = "T")
+public abstract class AbstractImmutableUnionFind<T> implements UnionFind<T> {
```
baierd, line 20: Why is there no interface `ImmutableUnionFind` that this class implements?

src/org/sosy_lab/common/collect/union_find/AbstractImmutableUnionFind.java
```diff
@@ -0,0 +1,28 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package org.sosy_lab.common.collect.union_find;
+
+import com.google.errorprone.annotations.DoNotCall;
+import com.google.errorprone.annotations.Immutable;
+
+/**
+ * An abstract class for immutable Union-Find implementations.
```
baierd, line 15: This description is not helpful. It doesn't tell users or developers anything about what to expect from this class or its inheritors.

src/org/sosy_lab/common/collect/union_find/AbstractImmutableSortedUnionFind.java
```diff
@@ -0,0 +1,20 @@
...
+
+import com.google.errorprone.annotations.Immutable;
+
+/**
+ * An abstract class for sorted immutable Union-Find implementations.
+ *
+ * @param <T> type of elements added to the Union-Find. Must be comparable.
+ */
+@Immutable(containerOf = "T")
+public abstract class AbstractImmutableSortedUnionFind<T extends Comparable<T>>
```
baierd, line 19: Why not use a interface?

src/org/sosy_lab/common/collect/union_find/AbstractImmutableParentPointerTreeBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+  // build() returns an immutable union-find that contains a copy of this union-find's map
+  @SuppressWarnings("Immutable")
+  final ParentPointerTreeUnionFind<T> unionFind;
+
+  protected AbstractImmutableParentPointerTreeBuilder(UnionType pUnionType) {
+    unionFind = new ParentPointerTreeUnionFind<>(pUnionType);
+  }
+
+  @CanIgnoreReturnValue
+  public AbstractImmutableParentPointerTreeBuilder<T> union(T pE1, T pE2) {
```
baierd, line 45: Most of the documentation in this class is missing,

src/org/sosy_lab/common/collect/union_find/AbstractImmutableParentPointerTreeBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+package org.sosy_lab.common.collect.union_find;
+
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import com.google.errorprone.annotations.Immutable;
+import java.util.Collection;
+import java.util.Set;
+import org.sosy_lab.common.collect.union_find.ParentPointerTreeUnionFind.UnionType;
+
+/**
+ * Abstract builder class which first collects the data in a mutable Union-Find and converts it to
```
baierd, line 18: What is "the data"? What form is the data supposed to have? What is allowed, what is disallowed? Are all "data" always collected into a single set inside the Union-Find?

src/org/sosy_lab/common/collect/union_find/NonRootNode.java
```diff
@@ -0,0 +1,28 @@
...
+
+package org.sosy_lab.common.collect.union_find;
+
+/**
+ * An implementation of {@link AbstractTreeNode} resulting in nodes that can only be used as
+ * non-root nodes but not as root nodes.
+ *
+ * @param <T> type of elements each node holds as value
+ */
+public final class NonRootNode<T> extends AbstractTreeNode<T> {
```
baierd, line 17: This class is equal to `AbstractTreeNode` and does not make sense. There is no distinction between nodes and abstract nodes at all, hence why not merge them?

src/org/sosy_lab/common/collect/union_find/AbstractTreeNode.java
```diff
@@ -0,0 +1,55 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package org.sosy_lab.common.collect.union_find;
+
+/**
+ * An abstract class of nodes from which a simple parent pointer tree can be built.
+ *
+ * @param <T> type of elements each node holds as value
+ */
+public abstract class AbstractTreeNode<T> {
```
baierd: How to discern leafs and roots in this implementation?

src/org/sosy_lab/common/collect/union_find/RootNode.java
```diff
@@ -0,0 +1,66 @@
...
+ * An implementation of {@link AbstractTreeNode} resulting in nodes that can be used as non-root
+ * nodes or root nodes. Their primary intended use is as root nodes. Rank describes the maximum
+ * height of the tree of this node and its child nodes (i.e. without path compression). Size
+ * describes the total number of elements in the tree represented by this root node.
+ *
+ * @param <T> type of elements each node holds as value
+ */
+public final class RootNode<T> extends AbstractTreeNode<T> {
+
+  private int rank;
```
baierd, line 21: Having both, rank and size, in a single class makes no sense. You are overloading the implementation of one with the implementation of the other. This makes code less readable, understandable, and maintainable, while at the same time doubling the space required for this class.

This should be split into 2 sub-classes.

src/org/sosy_lab/common/collect/union_find/RootNode.java
```diff
@@ -0,0 +1,66 @@
...
+//
+// SPDX-FileCopyrightText: 2026 Dirk Beyer <https://www.sosy-lab.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package org.sosy_lab.common.collect.union_find;
+
+/**
+ * An implementation of {@link AbstractTreeNode} resulting in nodes that can be used as non-root
+ * nodes or root nodes. Their primary intended use is as root nodes. Rank describes the maximum
```
baierd, line 13: Your class is named `RootNode`, yet you claim this can be used as non-root nodes here. That does not make sense.

src/org/sosy_lab/common/collect/union_find/AbstractTreeNode.java
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Constructor for a root node. The parent variable points to itself, thus indicating this is a
+   * root node. If appended to another tree, parent can be reallocated to the new parent node, while
+   * the current node simply functions as a non-root node from then on.
+   *
+   * @param pValue element to be stored in the node
+   */
+  protected AbstractTreeNode(T pValue) {
+    parent = this;
```
baierd, line 29: This may lead to infinite recursion when traversing along `getParent()` due to the lack of documentation. Why not set this to `null` or better, use an `Optional`?

src/org/sosy_lab/common/collect/union_find/AbstractTreeNode.java
```diff
@@ -0,0 +1,55 @@
...
+// https://github.com/sosy-lab/java-common-lib
+//
+// SPDX-FileCopyrightText: 2026 Dirk Beyer <https://www.sosy-lab.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package org.sosy_lab.common.collect.union_find;
+
+/**
+ * An abstract class of nodes from which a simple parent pointer tree can be built.
```
baierd, line 12: How to discern leafs and roots in this implementation? The documentation of how to use this to traverse the tree is completely missing.

src/org/sosy_lab/common/collect/union_find/AbstractTreeNode.java
baierd, line 16: `AbstractTreeNode` does not convey any information about the type of tree. Since you have a non-default tree here, the name should reflect this.

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+  }
+
+  // like find; returns node of canonical element of the set pE belongs to, not value
+  private RootNode<T> findNode(T pE) {
+
+    Preconditions.checkNotNull(pE);
+
+    List<AbstractTreeNode<T>> toBeCompressed = new ArrayList<>();
+    @Var AbstractTreeNode<T> node = allNodes.get(pE);
```
baierd, line 281: The naming of the two control variables (`node` and `parent`) could be improved, as current the names do not reflect their usage. Alternatively, a comment would help.

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+
+      while (!node.equals(parent)) {
+        toBeCompressed.add(node);
+        node = parent;
+        parent = node.getParent();
+      }
+
+      for (AbstractTreeNode<T> current : toBeCompressed) {
+
+        current.setParent(parent);
```
baierd, line 294: It would make sense to refactor path compression into a dedicated subclass, so that you retain both, the base version without this optimization, and the optimization for your evaluation.

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+  @Override
+  public void union(T pE1, T pE2) {
+
+    Preconditions.checkNotNull(pE1);
+    Preconditions.checkNotNull(pE2);
+
+    if (pE1.equals(pE2)) {
+      addElementAsNewSet(pE1);
+    } else {
+      if (contains(pE1)) {
```
baierd, line 106: There are multiple nested contains checks and `get()` operations for the same element(s). This can be improved.

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+
+      for (AbstractTreeNode<T> current : toBeCompressed) {
+
+        current.setParent(parent);
+      }
+
+      return (RootNode<T>) parent;
+    }
+
+    throw new IllegalArgumentException("Element not contained.");
```
baierd, line 300: If there is no element, you should throw a `NoSuchElementException`. There are multiple instances of this.

Furthermore, failing early is preferred. Hence it makes sense to check this and throw immediately after `get()`, since there is also no documentation describing whether the element may be null (i.e. `@Nullable`). Otherwise someone might add code in between.

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+
+    int size1 = pCanon1.getSize();
+    int size2 = pCanon2.getSize();
+
+    if (size1 > size2) {
+      pCanon2.setParent(pCanon1);
+      pCanon1.incrementSizeBy(pCanon2.getSize());
+    } else {
+      pCanon1.setParent(pCanon2);
+      pCanon2.incrementSizeBy(pCanon1.getSize());
```
baierd, line 251: Why not use the sizes above?

src/org/sosy_lab/common/collect/union_find/RootNode.java
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Constructor for a root node. The parent variable points to itself, thus indicating this is a
+   * root node. If appended to another tree, parent can be reallocated to the new parent node, while
+   * the current node simply functions as a non-root node from then on. In the beginning, rank is 0
+   * and size is 1.
+   *
+   * @param pValue element to be stored in the node
+   */
+  public RootNode(T pValue) {
```
baierd, line 32: This is not guaranteed to be a root node, as the parent is mutable. Switching this to a leaf is trivial, and the class in variants (i.e. size/rank accuracy) is no longer guaranteed.

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+    @Var T canon = null;
+
+    for (T current : pSet) {
+
+      if (canon == null) {
+        canon = current;
+        addElementAsNewSet(canon);
+      }
+
+      addElementToExistingSet(current, findNode(canon));
```
baierd, line 186: Why not simply use `union()`?

`findNode(canon)` always returns the same element.

src/org/sosy_lab/common/collect/union_find/AbstractGenericUnionFind.java
```diff
@@ -0,0 +1,179 @@
...
+
+    assert set1 != null;
+    assert set2 != null;
+
+    int size1 = set1.size();
+    int size2 = set2.size();
+
+    if (size1 > size2) {
+      set1.addAll(set2);
+      assert mapOfSets.remove(pE2, set2);
```
baierd, line 139: Assertions are not evaluated if the correct flags are not set and should never contain actual code.

src/org/sosy_lab/common/collect/union_find/PersistentParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,306 @@
...
+    return this;
+  }
+
+  private PersistentUnionFind<T> addElementAsNewSetAndCopy(T pE) {
+
+    if (!contains(pE)) {
+      Map<T, T> updatedNodesToParents = new HashMap<>(mapOfNodesToParents);
+      updatedNodesToParents.put(pE, pE);
+
+      Map<T, Integer> updatedRootsToWeights = new HashMap<>(mapOfRootsToWeights);
```
baierd, line 191: The class is described as `provides cheap copy-and-write operations.`.
But these `HashMap`s copy all elements, resulting in O(n) copies. Structural sharing should be used as far as possible in persistent implementations.

src/org/sosy_lab/common/collect/union_find/PersistentParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,306 @@
...
+
+  // map only used internally and never mutated after creation
+  // TODO convert to ImmutableMap once Guava upgraded to 31.1 and buildKeepingLast() available
+  @SuppressWarnings("Immutable")
+  private final Map<T, Integer> mapOfRootsToWeights;
+
+  private final UnionType unionType;
+
+  private PersistentParentPointerTreeUnionFind(UnionType pUnionType) {
+    mapOfNodesToParents = new HashMap<>();
```
baierd, line 50: `HashMap`s do not provide consistent iteration order, nor immutability, persistence. Hence the entire data-structure needs to be deep-copied every time a new element is added. Since there is also no structural sharing, this is no copy-efficient solution. I recommend using our `PersistentMap` implementation.

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+ */
+public class ParentPointerTreeUnionFind<T> implements UnionFind<T> {
+
+  public enum UnionType {
+    UNION_BY_RANK,
+    UNION_BY_SIZE
+  }
+
+  protected final Map<T, AbstractTreeNode<T>> allNodes;
+  private final UnionType unionType;
```
baierd, line 39: Same problem as with the nodes. The distinct union optimizations should be in their own subclasses.

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+ * @param <T> type of elements added to the Union-Find.
+ */
+public class ParentPointerTreeUnionFind<T> implements UnionFind<T> {
+
+  public enum UnionType {
+    UNION_BY_RANK,
+    UNION_BY_SIZE
+  }
+
+  protected final Map<T, AbstractTreeNode<T>> allNodes;
```
baierd, line 38: Why is this protected? Please use a getter instead.

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+  private final UnionType unionType;
+
+  /**
+   * Creates an empty instance.
+   *
+   * @param pUnionType type of union to be performed for all unions on this instance
+   */
+  public ParentPointerTreeUnionFind(UnionType pUnionType) {
+    allNodes = new HashMap<>();
```
baierd, line 47: `HashMap`s don't have a consistent iteration order. Please consider replacing them accordingly. (This should be done in general for all data-structures that do not provide a consistent iteration order in all cases.)

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+   * @param pE element for which set is to be found
+   * @return canonical element of the found set
+   * @throws IllegalArgumentException if element is not contained in any subset
+   */
+  @Override
+  public T find(T pE) {
+
+    Preconditions.checkNotNull(pE);
+
+    List<AbstractTreeNode<T>> toBeCompressed = new ArrayList<>();
```
baierd, line 64: Why not use `findNode()`?

src/org/sosy_lab/common/collect/union_find/AbstractImmutableParentPointerTreeBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+  @CanIgnoreReturnValue
+  public AbstractImmutableParentPointerTreeBuilder<T> union(T pE1, T pE2) {
+
+    unionFind.union(pE1, pE2);
+
+    return this;
+  }
+
+  @CanIgnoreReturnValue
+  public AbstractImmutableParentPointerTreeBuilder<T> add(Set<T> pSet) {
```
baierd, line 56: Missing documentation.

What happens if this set contains elements that are already in the union-find?

src/org/sosy_lab/common/collect/union_find/PersistentSortedUnionFind.java
```diff
@@ -0,0 +1,46 @@
...
+ * provides cheap copy-and-write operations. Thus, all write operations ({@link #union(Comparable,
+ * Comparable)}) will not modify the current instance, but return a new instance instead.
+ *
+ * <p>All modifying operations inherited from {@link SortedUnionFind} are not supported and will
+ * always throw {@link UnsupportedOperationException}.
+ *
+ * @param <T> The type of values.
+ */
+@Immutable(containerOf = "T")
+public interface PersistentSortedUnionFind<T extends Comparable<T>> extends SortedUnionFind<T> {
```
baierd: This should be a subtype of `PersistentUnionFind`.

Also, the type bounds are restrictive and disallow naturally ordered elements if the ordering is defined in a superclass. Please change this to `T extends Comparable<? super T>`.

src/org/sosy_lab/common/collect/union_find/tests/ImmutableParentPointerTreeUnionFindTest.java
```diff
@@ -0,0 +1,399 @@
...
+import org.sosy_lab.common.collect.union_find.AbstractImmutableParentPointerTreeBuilder;
+import org.sosy_lab.common.collect.union_find.AbstractImmutableUnionFind;
+import org.sosy_lab.common.collect.union_find.ImmutableParentPointerTreeUnionFind;
+import org.sosy_lab.common.collect.union_find.ImmutableSortedParentPointerTreeUnionFind;
+import org.sosy_lab.common.collect.union_find.ParentPointerTreeUnionFind.UnionType;
+import org.sosy_lab.common.collect.union_find.UnionFind;
+
+public class ImmutableParentPointerTreeUnionFindTest {
+
+  private final int[] simpleUnionArgs = new int[] {0, 1};
```
baierd, line 27: Why not just use the values? Why does this exist at all?

src/org/sosy_lab/common/collect/union_find/AbstractImmutableParentPointerTreeBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param <T> type of elements added to the Union-Find
+ */
+@Immutable(containerOf = "T")
+public abstract class AbstractImmutableParentPointerTreeBuilder<T> {
+
+  // union-find not immutable but only used internally and never mutated passed outward
+  // build() returns an immutable union-find that contains a copy of this union-find's map
+  @SuppressWarnings("Immutable")
+  final ParentPointerTreeUnionFind<T> unionFind;
```
baierd, line 32: Missing privacy modifier.

baierd (changes requested):
The implementation currently lacks documentation in many places and has outdated/wrong documentation in other places.

Class structure needs to be improved urgently, as it blocks several default methods/concepts (e.g. `Builder`) from being implemented correctly, and also mixes multiple distinct implementations. In other cases, e.g. `find()`, there is no need for having multiple distinct implementations, yet there are multiple 1:1 duplicates.

Several immutable or persistent classes have `SuppressWarnings("Immutable")` annotations that hide real problems. The biggest of which is that the ImmutableUnionFind is not immutable, since the tree is not immutable. Modifying nodes modifies other copies of the same tree. This is also not covered by tests.

Furthermore, there are code duplicates in multiple places that need to be reduced (I did not mark all, but they should be obvious when working on the marked cases). 

While tests cover default behavior, they miss type specifics quite often. But they too suffer from a structural problem, in that there is lots of duplicate code or test cases (`ParentPointerTreeUnionFindByRankTest` and `ParentPointerTreeUnionFindBySizeTest` are more or less direct duplicates), type inheritance is not used to test multiple union-finds at the same time etc. This would help with all test problems!
Please refactor the tests into abstract, generic, parameterized tests.
Also, don't modify test class variables in multiple places before using them (e.g. `SortedUnionFindTest`), as it becomes really hard to follow this, leading to tests easily being based on wrong assumptions.
Several tests for sorted variants test unsorted variants (e.g. `PersistentParentPointerTreeUnionFind` is tested instead of `PersistentSortedParentPointerTreeUnionFind`).
The tests for immutable versions don't test mutations of shared existing nodes (see above).
Tests are more or often then not undocumented. This is no problem for trivial tests, but more complex tests need to be documented. 
Optimizations are effectively untested.

baierd (commented):
Furthermore, the node implementations do not define `equals()` and `hashcode()`. This is not wrong, since nodes are not re-introduced in the implementations. But there are two problems due to this:
- the code is fragile, since expression like this `while (!node.equals(parent))` depend on reference equality implicitly now. Making this explicit helps code readability and future developers and maintainers. Switching to explicit reference equality should be considered.
- sorted implementations have a equals/compareTo mismatch. Example:
```
new BigInteger("1.0").equals(new BigInteger("1.00"))   // false
new BigInteger("1.0").compareTo(new BigInteger("1.00")) // 0, i.e. equal
```
I am currently unsure whether this can lead to problems in the current implementation.
Defining `equals()` and `hashcode()`, avoids these problems.

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+   *
+   * @param pUnionType type of union to be performed for all unions on this instance
+   */
+  public ParentPointerTreeUnionFind(UnionType pUnionType) {
+    allNodes = new HashMap<>();
+    unionType = pUnionType;
+  }
+
+  /**
+   * Returns the canonical element of the set containing the provided element. Applies path
```
baierd: Path compression is no concern for users when using this method, hence it should not be mentioned.

src/org/sosy_lab/common/collect/union_find/ParentPointerTreeUnionFind.java
```diff
@@ -0,0 +1,302 @@
...
+      }
+
+      return parent.getValue();
+    }
+
+    throw new IllegalArgumentException("Element not contained.");
+  }
+
+  /**
+   * Merges the sets represented by the two input values according to standard Union-Find behaviour.
```
baierd: What is standard union-find behavior? 
There are many implementations and flavors of this data-structure, which standard are you referring to?

It is best to just describe what is happening directly.

src/org/sosy_lab/common/collect/union_find/ImmutableParentPointerTreeUnionFind.java
baierd: Example:
```
builder.union(0, 1);
builder.union(2, 3);

AbstractImmutableUnionFind<Integer> beforeMerge = builder.build();

builder.union(1, 2); // This changes beforeMerge

AbstractImmutableUnionFind<Integer> afterMerge = builder.build(); // Shared UnionFind by accident
```

PR #65, opened 2026-09-07
Add a (copy-)efficient Persistent Stack Implementation

baierd:
We have a `PersistentStack` implementation in CPAchecker, but not in this library. It makes sense to move it here.

While doing so, the following was added:
- a interface for persistent stacks `PersistentStack`.
- a test class `PersistentLinkedStackTest`.
- a extended `PersistentLinkedStack` implementation that provides serialization through a proxy and an iterator + proper documentation.

Note: the current state is a hybrid of our `PersistentStack` in CPAchecker and `PersistentLinkedList` in this repo, but extended using LLM generated code. I reviewed the changes done by the LLM before opening this PR. I am unsure about the serialization though, so input would be appreciated.

### DIFF
--- a/src/org/sosy_lab/common/collect/PersistentLinkedStack.java
+++ b/src/org/sosy_lab/common/collect/PersistentLinkedStack.java
@@ -1,0 +1,273 @@
+// This file is part of SoSy-Lab Common,
+// a library of useful utilities:
+// https://github.com/sosy-lab/java-common-lib
+//
+// SPDX-FileCopyrightText: 2007-2026 Dirk Beyer <https://www.sosy-lab.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package org.sosy_lab.common.collect;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.collect.UnmodifiableIterator;
+import com.google.errorprone.annotations.Immutable;
+import com.google.errorprone.annotations.Var;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
+import java.io.Serial;
+import java.io.Serializable;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * A persistent stack. Pushes structurally share the complete previous stack, and pops return the
+ * existing tail without copying, while leaving the original stack unchanged. Thus {@link
+ * #pushAndCopy(Object)}, {@link #popAndCopy()}, {@link #peek()}, {@link #empty()}, {@link
+ * #isEmpty()}, and {@link #size()} run in O(1) time. Iteration, {@link #equals(Object)}, {@link
+ * #hashCode()}, and {@link #toString()} have O(n) worst-case stack-traversal overhead. When two
+ * equal-size {@link PersistentLinkedStack} instances share a tail, {@code equals} traverses only
+ * the k nodes preceding that tail and thus has O(k) traversal overhead. These bounds exclude work
+ * performed by element methods.
+ *
+ * <p>All structural state is final and correctly constructed, so the immutable stack structure may
+ * be accessed concurrently without synchronization. Publishing or updating a shared reference to a
+ * stack version still requires coordination, and compound updates require synchronization or an
+ * atomic operation. Each traversal has an independent iterator; a single iterator instance has no
+ * thread-safety guarantee. Values are stored by reference and are not copied or made immutable or
+ * thread-safe.
+ *
+ * <p>Null values are not supported.
+ *
+ * <p><b>Serialization:</b> Serialization and deserialization have O(n) stack-traversal overhead and
+ * use O(n) temporary memory, excluding the processing of element object graphs. A flattened proxy
+ * stores the logical values in top-to-bottom order instead of serializing the linked nodes. This
+ * avoids recursive traversal of long stacks and keeps node and cache fields out of the serialized
+ * form. Deserialization rejects null proxy data and rebuilds the stack from bottom to top while
+ * validating each value, thereby preserving order and restoring the canonical empty instance.
+ * Because each stack is flattened independently, distinct, structurally related stacks serialized
+ * together have their shared non-empty tails reconstructed independently; {@link #popAndCopy()} on
+ * a deserialized stack nevertheless returns its existing tail. Element object graphs must not
+ * contain references back to the containing stack because proxy replacement cannot restore such
+ * cycles. Persisted data remains readable only while the proxy and element serialized forms remain
+ * compatible.
+ *
+ * @param <T> the type of values
+ */
+@Immutable(containerOf = "T")
+public final class PersistentLinkedStack<T> implements PersistentStack<T> {
+
+  @Serial private static final long serialVersionUID = -4286928240765960519L;
+
+  private static final PersistentLinkedStack<?> EMPTY = new PersistentLinkedStack<>();
+
+  /** The top value, null exactly for the empty singleton. */
+  @SuppressWarnings("serial") // writeReplace prevents direct serialization of this field.
+  private final @Nullable T top;
+
+  /** The linked tail, null exactly for the empty singleton. */
+  private final @Nullable PersistentLinkedStack<T> tail;
+
+  /**
+   * The size, cached for O(1) access. It is zero exactly for the empty stack and otherwise equals
+   * {@code tail.size + 1}. The cache is one logical 4-byte {@code int}; its actual footprint
+   * depends on JVM object layout and alignment. It often fits into padding with compressed
+   * references and 8-byte alignment, but may add an alignment unit otherwise.
+   */
+  private final int size;
+
+  private PersistentLinkedStack() {
+    top = null;
+    tail = null;
+    size = 0;
+  }
+
+  private PersistentLinkedStack(T pTop, PersistentLinkedStack<T> pTail) {
+    top = checkNotNull(pTop);
+    tail = checkNotNull(pTail);
+    size = pTail.size + 1;
+  }
+
+  /** Returns an empty stack. */
+  @SuppressWarnings("unchecked")
+  public static <T> PersistentLinkedStack<T> of() {
+    return (PersistentLinkedStack<T>) EMPTY;
+  }
+
+  /**
+   * Returns a stack containing {@code value}.
+   *
+   * @throws NullPointerException if {@code value} is null
+   */
+  public static <T> PersistentLinkedStack<T> of(T value) {
+    return new PersistentLinkedStack<>(value, PersistentLinkedStack.of());
+  }
+
+  @Override
+  public PersistentLinkedStack<T> pushAndCopy(T value) {
+    return new PersistentLinkedStack<>(value, this);
+  }
+
+  @Override
+  public PersistentLinkedStack<T> popAndCopy() {
+    if (isEmpty()) {
+      throw new NoSuchElementException();
+    }
+    return checkNotNull(tail);
+  }
+
+  @Override
+  public T peek() {
+    if (isEmpty()) {
+      throw new NoSuchElementException();
+    }
+    return checkNotNull(top);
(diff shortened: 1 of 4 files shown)


src/org/sosy_lab/common/collect/PersistentStack.java
```diff
@@ -0,0 +1,85 @@
...
+ * thread-safety guarantee is made for iterator instances.
+ *
+ * <p>Values are stored by reference: they are not copied or made immutable or thread-safe. Changes
+ * to mutable values can affect equality and hash codes. Operations that depend on values also
+ * depend on their thread safety.
+ *
+ * @param <T> The type of values.
+ */
+@Immutable(containerOf = "T")
+public interface PersistentStack<T> extends Iterable<T>, Serializable {
```
PhilippWendler, line 51: I would think it is better to not extend `Iterable` in `PersistentStack`. I personally find it always confusing what the iteration order of a stack is, and would never be sure without looking it up, also because of inconsistency between related classes: For `PersistentStack`, the only meaningful iteration order is top-down (last element pushed is first). The (de facto deprecated and unused, but still existing) `java.util.Stack` has a bottom-up iteration order, because it is based on a list and pushes to the end. The standard stack in Java (`java.util.Deque`) iterates top-down if you use `push()`, but bottom-up if you use `add()`.

We can instead provide a O(1) method `asReverseIterable` or so that returns an `Iterable` (or likely even a `SequencedCollection` if not a `List`) view. The name would make it clear that the iteration order is the reverse of the addition order. Or maybe `asTopDownIterable()` even?
Additionally we can provide a O(n) `copyToList()` that returns the elements in addition (bottom-up) order.

src/org/sosy_lab/common/collect/PersistentStack.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * <p>After a stack reference has been made visible to other threads through synchronization, a
+ * {@code volatile} field, or a concurrency utility, its immutable structure may be accessed
+ * concurrently. Such coordination is still required to publish or update a shared reference to a
+ * stack version, and compound updates require synchronization or an atomic operation. No
+ * thread-safety guarantee is made for iterator instances.
+ *
+ * <p>Values are stored by reference: they are not copied or made immutable or thread-safe. Changes
+ * to mutable values can affect equality and hash codes. Operations that depend on values also
+ * depend on their thread safety.
```
PhilippWendler, line 46: Sure, this is all true, but it is also the same for all such collections and pretty standard. I would shorten to just say something like "Serializable if all elements are serializable" and "thread safe including its collection views, but no guarantees for the iterator".

src/org/sosy_lab/common/collect/PersistentLinkedStackTest.java
```diff
@@ -0,0 +1,248 @@
...
+
+import com.google.common.testing.EqualsTester;
+import com.google.common.testing.SerializableTester;
+import com.google.errorprone.annotations.Var;
+import java.math.BigInteger;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import org.junit.Test;
+
+public class PersistentLinkedStackTest {
```
PhilippWendler, line 22: Please add tests based on Guava's testlib like we have for `PathCopyingPersistentTreeMap`. These are much more comprehensive than what we would think of.

src/org/sosy_lab/common/collect/PersistentStack.java
PhilippWendler, line 51: What is missing in this interface is a definition of `equals` and `hashCode`. Typical data structures define these precisely such that even different implementations of the same interface are guaranteed to compare equal. The current `equals` and `hashCode` implementations would not achieve this, for example.

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
```diff
@@ -0,0 +1,273 @@
...
+        if (value == null) {
+          throw new InvalidObjectException("Stack values must not contain null");
+        }
+        stack = stack.pushAndCopy(value);
+      }
+      return stack;
+    }
+  }
+
+  private static final class StackIterator<T> extends UnmodifiableIterator<T> {
```
PhilippWendler, line 249: Could use [`AbstractIterator`](https://github.com/google/guava/wiki/CollectionHelpersExplained#abstractiterator).

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
```diff
@@ -0,0 +1,273 @@
...
+
+    private @Nullable PersistentLinkedStack<T> stack;
+
+    private StackIterator(PersistentLinkedStack<T> pStack) {
+      stack = pStack;
+    }
+
+    @Override
+    public boolean hasNext() {
+      return stack != null && !stack.isEmpty();
```
PhilippWendler, line 259: Only one of these conditions should be needed. If is confusing if there are two possible different states of the iterator that both mean "end of iteration".

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+    return hashCode;
+  }
+
+  /**
+   * Returns the values in top-to-bottom order, separated by {@code ", "} and enclosed in square
+   * brackets: {@code [top, ..., bottom]}. The empty stack is represented as {@code []}.
+   */
+  @Override
+  public String toString() {
```
PhilippWendler, line 188: Can be implemented trivially using `Joiner` or `Collectors.joining()`.

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
```diff
@@ -0,0 +1,273 @@
...
+ * together have their shared non-empty tails reconstructed independently; {@link #popAndCopy()} on
+ * a deserialized stack nevertheless returns its existing tail. Element object graphs must not
+ * contain references back to the containing stack because proxy replacement cannot restore such
+ * cycles. Persisted data remains readable only while the proxy and element serialized forms remain
+ * compatible.
+ *
+ * @param <T> the type of values
+ */
+@Immutable(containerOf = "T")
+public final class PersistentLinkedStack<T> implements PersistentStack<T> {
```
PhilippWendler, line 60: We also want factory methods, which then can be used for deserialization.

And for example a `Collector` implementation is also missing.

PhilippWendler:
We can consider adding some bulk operations, for example a method to get the bottom n elements, similar to `sublist()`, and maybe more.

baierd:
> We can consider adding some bulk operations, for example a method to get the bottom n elements, similar to `sublist()`, and maybe more.

I added a implementation for this.

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
baierd: I added them.

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
baierd: Done.

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
baierd: Good spot, this was left over from a previous iteration. I cleaned it up.

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
baierd: Oh thats much better! I refactored it into a `AbstractIterator`.

src/org/sosy_lab/common/collect/PersistentStack.java
baierd: I added it. Please take a closer look, i might have over-interpreted the request a little.

src/org/sosy_lab/common/collect/PersistentStack.java
baierd: I shortened it a little.

src/org/sosy_lab/common/collect/PersistentStack.java
baierd: Implemented as proposed.

src/org/sosy_lab/common/collect/PersistentLinkedStackTest.java
baierd: I tried following it and refactored the tests. Could you please take a look whether this is sufficient?

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
```diff
@@ -0,0 +1,316 @@
...
+      result = result.pushAndCopy(value);
+    }
+    return result;
+  }
+
+  /** Returns a collector that pushes elements in encounter order, with the last element on top. */
+  @SuppressWarnings("NoFunctionalReturnType")
+  public static <T> Collector<T, ?, PersistentLinkedStack<T>> toPersistentLinkedStack() {
+    return Collectors.collectingAndThen(
+        ImmutableList.<T>toImmutableList(), PersistentLinkedStack::copyOf);
```
PhilippWendler, line 246: I don't think we want to first copy everything into a list and then into a stack. That is what users could easily do without our collector.

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  @Override
+  public Iterable<T> asTopDownIterable() {
+    return () -> new StackIterator<>(this);
+  }
+
+  @Override
+  public ImmutableList<T> copyToList() {
+    return ImmutableList.copyOf(asTopDownIterable()).reverse();
```
PhilippWendler, line 154: We could use a `Builder` with the known size to make this more efficient.

src/org/sosy_lab/common/collect/PersistentStack.java
```diff
@@ -0,0 +1,116 @@
...
+  boolean isEmpty();
+
+  /** Returns the number of values in this stack. */
+  int size();
+
+  /**
+   * Returns an unmodifiable top-to-bottom view in O(1) time. Each iterator traverses this stack
+   * version independently.
+   */
+  Iterable<T> asTopDownIterable();
```
PhilippWendler, line 78: If we can return a `Collection` here it could be more convenient for callers. In principle `SequencedCollection` would be even better semantically, but it only adds `reversed()`, which we can hardly support, and some mutation methods. So there is little point in it from the API point of view.

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
```diff
@@ -0,0 +1,316 @@
...
+      @Var PersistentLinkedStack<Object> stack = PersistentLinkedStack.of();
+      // The serialized order is top-to-bottom; push in the opposite direction.
+      for (@Var int index = serializedValues.length - 1; index >= 0; index--) {
+        @Nullable Object value = serializedValues[index];
+        if (value == null) {
+          throw new InvalidObjectException("Stack values must not contain null");
+        }
+        stack = stack.pushAndCopy(value);
+      }
+      return stack;
```
PhilippWendler, line 294: We can replaced this with something based on `copyOf(Arrays.asList(values).reversed())`.

src/org/sosy_lab/common/collect/PersistentLinkedStackTest.java
PhilippWendler: I don't think testing the result of `copyToList()` is useful. This would test the creation of the stack, the `copyToList()` call and then all the features of the returned list implementation and its views. But it will not test any of the other `PersistentLinkedStack` features and methods.

src/org/sosy_lab/common/collect/PersistentLinkedStack.java
PhilippWendler: Actually, we should keep the code here as is, but it is a good argument for `asTopDownIterable()` to return a `Collection`, because then `copyof()` will automatically use the correct size.

PR #66, opened 2026-09-14
Reuse objects when inserting identical keys/values in PathCopyingPersistentTreeMap

PhilippWendler:
This PR affects cases where objects are inserted into a `PathCopyingPersistentTreeMap` and the value is identical (as defined by `==`) as the old value for the same key. Now we reuse the old key object (even if it is not the same instance as the newly given key object) and then can reuse also the same `Node` and map objects, saving a little bit of memory for the cost of a single `==` comparison on re-inserting with the same key.

Map keys need to have well-defined equality (here actually `compareTo()` because it is a sorted map), so code that would rely on getting a specific instance of a key and breaks with other instances of an equal key is broken anyway and relies on assumptions that were never guaranteed.

Note that we do not pose any assumptions on value equality, as it is legal to put arbitrary objects as values into maps, even if they have broken/non-standard equals, and expect to get the same instance back as put last. This is for example necessary for safely putting mutable data structures like lists as values into a map.

The new behavior is the same as for JDK's `HashMap` and `TreeMap`, and there are tests for confirming this.

@lembergerth @baierd Might one of you want to review this?


lembergerth:
I think @baierd is quite busy at the moment; @baierd if that's true and you don't insist on reviewing this PR, I can do it (but I don't want to do double work)

lembergerth:
~~@baierd just offered to do it off-line; sorry for the back-and-forth~~ (based on a misunderstanding, I'll do it)

src/org/sosy_lab/common/collect/PathCopyingPersistentTreeMap.java
```diff
@@ -555,7 +563,9 @@ private static <K extends Comparable<? super K>, V> Node<K, V> putAndCopy0(
       current = current.withRightChild(newRight);
 
     } else {
-      current = new Node<>(key, value, current.left, current.right, current.getColor());
+      // This always keeps the old key object (if it compares equal but is not the same instance),
+      // but uses the new value object. This is not strictly necessary but what JDK maps do as well.
```
lembergerth: _suggestion_: Explain in the comment the advantage of reusing 'current' instead of always creating a new Node object.

The comment itself left me wondering why this should be done just because the JDK does it. I only understood through your PR description that this is to avoid unnecessary object creations / to save memory.



```suggestion
      // but uses the new value object. This is not strictly necessary but saves memory by reusing the Node object. This is what JDK maps do as well.
```

lembergerth: _suggestion (non-blocking)_: Explain in the comment the advantage of reusing the key.

The comment itself left me wondering why this should be done just because the JDK does it. I only understood through your PR description that this is to avoid unnecessary object creations / to save memory.

I'm also not sure I understand whether `but is not the same instance` is relevant.
At this branch, it is already certain that the old and the new key object are equal (`key.compareTo(current.getKey()) == 0`).
Doesn't this also happen when old and new key are the same instance, i.e., `key == current.getKey()`?

Comment from what I understood:

```suggestion
      // This always keeps the old (equal) key object. This has useful implications:
      // Because we reuse the old key object, the key instance does not change and potential `==` comparisons on the key at other locations still work successfully.
      // We do always use the new value object; but in case that the new value object is identical to the old value object, we can reuse the whole existing Node object. This also enables `==` comparisons to succeed and saves some memory.
      // This behavior also matches what JDK maps do.
```

Of course I'm not sure the explanation is fully correct; but I would suggest to clarify the comment somehow.

src/org/sosy_lab/common/collect/PathCopyingPersistentTreeMap.java
lembergerth, line 542: _nitpick_: Maybe the JavaDoc of [`PersistentMap#putAndCopy`](https://github.com/sosy-lab/java-common-lib/blob/main/src/org/sosy_lab/common/collect/PersistentMap.java#L38-L40) should be updated:

```java
/** Replacement for {{@link #put(Object, Object)} that returns a fresh instance. */
  @CheckReturnValue
  PersistentMap<K, V> putAndCopy(@CompatibleWith("K") K key, @CompatibleWith("V") V value);
```

In the case that `key` is equal to the existing key and `value` is identical to the existing value,
this method may return the same instance as before, not a fresh one.

lembergerth (approved):

PhilippWendler:
Thanks, you are right. I updated both the internal and external comments (in a few places). I didn't specifically add the sentence about returned objects being potentially identical, I think this is not surprising for a persistent data structure. But I made sure that we do not actually hint at the opposite.

PR #67, opened 2026-09-19
Replace persistent linked list impl with persistent stack

baierd:
This PR replaces the current `PersistentLinkedList` implementation with our new `PersistentStack` in the background as a delegate, as they are conceptually the same.

This should only be merged strictly after #65.


PhilippWendler:
Will this have `equals()` and `hashCode()` implementations that conform to the `List` requirements?

PhilippWendler:
What is the goal of this? Why should one use `PersistentLinkedList` instead of `PersistentLinkedStack`?

